PR #7, opened 2020-10-25
Initial implementation of collapsible question/answer

samliu10:
## **Summary**
- Closes issue #1 
- This pull request is for the basic implementation of the collapsible FAQ section using components from `reactstrap`.
- Each question and answer is contained in a `CollapsibleQuestion`.
- Questions are show in the card header, and the button on the right side controls the opening and closing of the card body to show the answer. The current placeholder toggle is a "+" that changes to "-" when the card body is opened.

**Before/After**
- This is the initial implementation of the FAQ section. 
- The FAQ section is currently implemented as a component on the main page, `App.tsx`. This branch does not have routing functionality at the moment, so this section will need to be included in the routed pages of the website later on.
- Each instance of component `CollapsibleQuestion` takes in 2 props: `question` and `answer`.

**Test Plan**
- Run `yarn start` in the directory `my-app` to see the FAQ section. 
- Placeholder data is used to display 3 questions and their corresponding answers.


ashritaraman (approved):
Looks good! Just a few minor things:

- I would give everything an explicit class name (so in CollapsibleQuestion.module.scss, I would add a class to the button instead of styling the Button component in general, just in case more buttons need to be added to the page in the future).
- In your pull request description, make sure to say which issue you are closing (so in this case, you should add the line 'Closes issue #1' so that the pull request can be tagged to the corresponding issue.

ashritaraman (changes requested):
Looks good! Just a few minor things:

- I would give everything an explicit class name (so in CollapsibleQuestion.module.scss, I would add a class to the button instead of styling the Button component in general, just in case more buttons need to be added to the page in the future).

- In your pull request description, make sure to say which issue you are closing (so in this case, you should add the line 'Closes issue #1' so that the pull request can be tagged to the corresponding issue.

ashritaraman (approved):

menaattia (commented):
Looks good but maybe get rid of the extra lines in between components in the CollapsableQuestion.tsx file

PR #8, opened 2020-10-26
Initial Homepage and Routes

menaattia:
# Summary
* Closes issue #2 
* Implemented the initial design for the homepage using material-ui components
* The individual components used for the homepage are in components/Home and the homepage is in pages/HomePage.tsx
* App.jsx has a route for the homepage using `react-router-dom`

## Test Plan
Run `yarn start` in my-app to see the homepage at the '/' route.

## Notes
Styling needs to be made responsive


ashritaraman (changes requested):
Looks good! A few things

- Resolve conflicts in package.json, App.tsx, and yarn.lock

- In your SearchBar.tsx files, you named one of your classes `inputInput` -- I would suggest using a more descriptive name

- Just a note (not a change): I see that you are using material-ui which is great, but we should probably use material-ui for all future branches if we want to go with a framework (this also depends on the final designs)

samliu10:
A few suggestions:
- Kinda similar to the feedback I got: make sure your components have specific class names, e.g. you can give buttons names.
- Since we're using scss for styling, I'd suggest moving your styling out of the components, i.e. take out `useStyles` in `SearchBar.tsx` and `NavBar.tsx` and make a `.module.scss` file(s) that will hold the styling. Then import the scss file(s) into those two components. Who knows- these components may get more complicated and could have a lot of styling, so keeping styling separate from the actual components themselves will keep the code cleaner and consistent among all of us developers.

ashritaraman (changes requested):
- Remove icon padding/surrounding background so that the icon doesn't shift to the right on the homepage

- Change name of 'my-app' folder to 'cu-apts'

- Remove Resources button

- Fix search bar padding so that it stays above the houses

- Create a folder under components called images, and place the home-icon logo in there

ashritaraman (approved):

Issue #5, opened 2020-10-24
Start thinking about backend implementation for FAQs


ashritaraman:
Initialize backend in firebase, think about how to structure the backend for FAQs (e.g. should the collections be organized by dropdown/by category? what attributes should be included?)

PR #13, opened 2020-11-10
Initial backend implementation of faq

menaattia:
## Summary
* Closes issue #5  
* This is the initial implementation of the FAQ backend.
* The express server is set up to take GET request to the route "/". The GET request returns all the faqs in the faq collection.
* firestore is initialized in firebase/firebase.ts
* connected the frontend and backend

# Test Plan
Run `npm install` in the backend folder. Then `npm run tsc` to compile TypeScript, followed by `node server.js` to start the server. In Postman, make a GET request to localhost:8080. The request should return faqs from the database. The data should also appear on the faq page on localhost:3000/faq by running `yarn start` in the frontend folder

# Notes
Anyone can currently read and write to the faq firestore collection. It is not an issue at the moment because there is no private or real data in the faqs collection, but the security rules need to be updated and `firestore-admin` needs to be set up.


samliu10:
Nice job with setting up the backend and connecting the backend to the frontend! It definitely looks a lot more realistic now that the frontend can be populated with data from Firebase. One thing though is that it seems like the FAQ sections are being populated in reverse order, e.g. Section 1 appears below Section 2 so we might want to take a look at that once we input real section titles.

ashritaraman (approved):

Issue #14, opened 2020-12-05
Implement inner home page (with listings)


ashritaraman:
Use the current homepage design as a starting point for implementing the new version of the home page with listings

kjasani111:
Closed with pull request #33

PR #18, opened 2020-12-12
Homepage listings page

kjasani111:
## Summary

- Closes issue #14 
- This pull request is for the basic implementation of the new homepage design that includes apartment listings.
### Before/After

- The background of the page is now a shade of white instead of an image.
- The search bar has been modified to match the new design requirements.
- A "Browse Renting Companies" section has been added, which can be populated with as many listings as required, and it will scroll horizontally through the listing cards so that 3 are showing at a time.
- Currently, the listing cards are populated with hardcoded data, but they have been made functional in a way that populating them with backend data will not be difficult.
### Testing

- Run `yarn install` in `frontend` to make sure all packages are installed.
- Run `yarn start` in `frontend` to view the React project in a local browser. It will open to the Home page, where you can see the implementation of the new design.


frontend/src/pages/HomePage.tsx
```diff
@@ -1,13 +1,21 @@
...
 
 const HomePage= (): ReactElement => {
     return <div className='Home'>
         <NavBar />   
-        <SearchBar />       
+        <SearchBar />   
+        <SectionDivider /> 
+        <div className="container-fluid">
+            <div className="row flex-row flex-nowrap pt-3">
+                <div className="col-4"><ApartmentCard address="117 Eddy St" company="Ithaca Renting Company" bedsAndBaths="5 Br | 2 B" amenities={false} price="$800" percentReview="90%" numReviews="5 Reviews"/></div>
```
ptwu, line 14: `react-bootstrap` is slightly different from doing bootstrap in vanilla JS. We shouldn't be using `<div>`s here, but should rather be using the column / grid components provided by the `react-bootstrap` library. See this link for how to do that: https://react-bootstrap.github.io/layout/grid/

In doing this, you should set each of these cards to be 12 columns wide (entire row) for mobile or small (`sm`) screens, and 4 columns for medium and up.

ptwu (changes requested):
Looks great! However, there are some things we need to fix before merging to main, namely mobile responsive support. Inspect element -> mobile view got me this with the apartment cards:

image

This is likely a result of how we're handling grids in `HomePage.tsx`. See my review comment there for some pointers. Also, we can do this in a future PR, but the apartment card data is currently hardcoded. We eventually want those to be dynamic and depend on some data.

PR #23, opened 2021-02-13
Setup eslint and prettier

Enochen:
### Summary 

ESLint is nice. Prettier is nice. I added them.

### Test Plan 

VSCode shows ESLint errors/warnings as expected.

You can also run commands like `yarn lint` and `yarn format:check` to check ESLint/Prettier errors respectively.

### Notes 

I ran auto fixes through the codebase but we can undo those in case they create merge conflicts for current PRs.


ptwu (dismissed):
looks good! we can add `yarn format` as a pre-commit hook to make matters even easier

ptwu (approved):
looks like I need to re-approve for this but the other changes were also ported to the branch in #24.

PR #24, opened 2021-02-14
ts-node fix

Enochen:
### Summary 

This diff moves the backend logic into a `src` folder and adjusts the `path` property in `tsconfig.json`.

This solves the issue of `ts-node` not working without running `tsc` beforehand (to generate the js files).

### Test Plan 

Running `yarn start` works.


ptwu (approved):
Looks good! Thanks for helping set up CI

PR #29, opened 2021-02-21
Faq spinner

samliu10:
### Summary 

This PR adds a spinner onto the FAQ page and includes some code reorganization.

- The spinner appears while the FAQ data is loading.
- Added minor css fixes to link colors upon hover in buttons on the main page and FAQ page. 
- Components shared between pages (search bar, spinner) are now located in a new folder called`utils`. 

### Test Plan

https://user-images.githubusercontent.com/57203639/108640390-51565d00-7467-11eb-81f3-f556e360d60f.mov

Start a dev server by running `yarn start` in the root directory. Right now, the data loads very quickly, so the spinner may not actually be visible. You can reverse the logic in `FAQPage.tsx` line 41 to show the spinner and see what it would look like.


frontend/src/App.scss
```diff
@@ -5,9 +5,17 @@ body {
...
 .links:hover {
   text-decoration: none;
+}
+
+.links {
   color: black;
 }
 
+.btn-outline-dark:hover > .links,
+.faq-button:hover > .links {
```
menaattia, line 15: Looks good! the spinner is working as expected. One small comment is I don't think the faq-button class exists; the faq button has the same btn-outline-dark class as the Reviews button, and when I ran the app without this line, the styles were still applied so you can remove this line.

ptwu:
Also, you might want to consider renaming the component to something like `ProgressSpinner` and into its own util components folder, since we will want to reuse this beyond just the FAQ page.

frontend/src/components/utils/ProgressSpinner.module.scss
```diff
@@ -0,0 +1,11 @@
+.container {
+  margin-top: 60px;
+  margin-left: auto;
+  margin-right: auto;
+  width: 60px;
+}
+
+.spinner {
+  height: 60px !important;
```
ptwu, line 9: Try to avoid `!important` when styling things. It usually means that you can change this CSS property somewhere else (maybe try the container div?)

ptwu (approved):

PR #30, opened 2021-02-24
Review Component

menaattia:
### Summary 

This pull request is the first step towards implementing reviews

- [x] implemented the Review component
- [x] implemented Rating component that takes in a rating out of 5 and fills in the hearts
- [x] added dummy reviews to pass as props to the Review component

### Test Plan 
Run `yarn start` from the workspace or frontend folder, and go to route `/landlords/landlord:id` using any dummy landlord id

review
review_mobile

### Notes 

- This implementation doesn't include the breakdown of ratings by maintenance, amenities, etc., but can be added later 
-  I included a name field in the dummy data, but hardcoded Anonymous as the name for now since we want all reviews to be anonymous 
- I didn't include the "Report Review" button because I'm not sure if we're including this in the MVP, but the button can be easily added

### DIFF
--- a/frontend/src/components/Review/Rating.tsx
+++ b/frontend/src/components/Review/Rating.tsx
@@ -1,0 +1,34 @@
+import React, { ReactElement } from 'react';
+import styles from './Rating.module.scss';
+import FavoriteIcon from '@material-ui/icons/Favorite';
+import FavoriteBorderIcon from '@material-ui/icons/FavoriteBorder';
+
+type Props = {
+  readonly rating: number;
+};
+const filled_hearts = (n: number) => {
+  let rows = [];
+  for (let i = 0; i < n; i++) {
+    rows.push(<FavoriteIcon fontSize="small" className={styles.rating} />);
+  }
+  return rows;
+};
+
+const empty_hearts = (n: number) => {
+  let rows = [];
+  for (let i = 0; i < n; i++) {
+    rows.push(<FavoriteBorderIcon fontSize="small" className={styles.rating} />);
+  }
+  return rows;
+};
+
+const Rating = ({ rating }: Props): ReactElement => {
+  return (
+    <div>
+      {filled_hearts(rating)}
+      {empty_hearts(5 - rating)}
+    </div>
+  );
+};
+
+export default Rating;
--- a/frontend/src/components/Review/Rating.tsx
+++ b/frontend/src/components/Review/Rating.tsx
@@ -6,11 +6,11 @@ import FavoriteBorderIcon from '@material-ui/icons/FavoriteBorder';
 type Props = {
   readonly rating: number;
 };
-const filled_hearts = (rating: number) => {
+const filledHearts = (rating: number) => {
   return [...Array(rating)].map(() => <FavoriteIcon fontSize="small" className={styles.rating} />);
 };
 
-const empty_hearts = (rating: number) => {
+const emptyHearts = (rating: number) => {
   return [...Array(5 - rating)].map(() => (
     <FavoriteBorderIcon fontSize="small" className={styles.rating} />
   ));
@@ -19,8 +19,8 @@ const empty_hearts = (rating: number) => {
 const Rating = ({ rating }: Props): ReactElement => {
   return (
     <div>
-      {filled_hearts(rating)}
-      {empty_hearts(rating)}
+      {filledHearts(rating)}
+      {emptyHearts(rating)}
     </div>
   );
 };
--- a/frontend/src/components/Review/Rating.tsx
+++ b/frontend/src/components/Review/Rating.tsx
@@ -6,27 +6,21 @@ import FavoriteBorderIcon from '@material-ui/icons/FavoriteBorder';
 type Props = {
   readonly rating: number;
 };
-const filled_hearts = (n: number) => {
-  let rows = [];
-  for (let i = 0; i < n; i++) {
-    rows.push(<FavoriteIcon fontSize="small" className={styles.rating} />);
-  }
-  return rows;
+const filled_hearts = (rating: number) => {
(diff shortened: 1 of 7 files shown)


frontend/src/pages/LandlordPage.tsx
```diff
@@ -1,12 +1,44 @@
...
+  },
+];
 
 const LandlordPage = (): ReactElement => {
   const { landlordId } = useParams<Record<string, string | undefined>>();
 
   return (
     <div>
       <h1>{`This is dummy text! My current landlordId is ${landlordId}`}</h1>
+      {reviews.map((review) => (
```
ptwu, line 35: Here, you can destructure `review` to be `{ date, text }`. This way, you don't have to do `review.date` and `review.text` and instead just use them as `date` and `text` respectively.

frontend/src/pages/LandlordPage.tsx
```diff
@@ -1,12 +1,44 @@
...
+        <Review
+          overall_rating={review.overall_rating}
+          date={review.date}
+          text={review.text}
+        ></Review>
```
ptwu, line 40: since this is an empty component, you can just do `<Review />` instead of having both an opening and closing tag

frontend/src/pages/LandlordPage.tsx
```diff
@@ -1,12 +1,44 @@
 import React, { ReactElement } from 'react';
 import { useParams } from 'react-router-dom';
+import Review from '../components/Review';
+
+let reviews = [
```
ptwu, line 5: this is constant data, so using a `const` is preferred.

frontend/src/components/Review/Rating.tsx
```diff
@@ -0,0 +1,34 @@
...
+    rows.push(<FavoriteBorderIcon fontSize="small" className={styles.rating} />);
+  }
+  return rows;
+};
+
+const Rating = ({ rating }: Props): ReactElement => {
+  return (
+    <div>
+      {filled_hearts(rating)}
+      {empty_hearts(5 - rating)}
```
ptwu, line 29: doing the computation when passing the number as an argument is bad practice - you should do it in the actual function instead

frontend/src/components/Review/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+  'June',
+  'July',
+  'Aug',
+  'Sept',
+  'Oct',
+  'Nov',
+  'Dec',
+];
+
+const date_to_string = (date: Date): string => {
```
ptwu, line 27: better to refactor this into a util script outside of the components folder so that we can reuse this if needed.

frontend/src/components/Review/Rating.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { ReactElement } from 'react';
+import styles from './Rating.module.scss';
+import FavoriteIcon from '@material-ui/icons/Favorite';
+import FavoriteBorderIcon from '@material-ui/icons/FavoriteBorder';
+
+type Props = {
+  readonly rating: number;
+};
+const filled_hearts = (n: number) => {
```
ptwu, line 9: `n` is not a very descriptive parameter name. It goes against what you typically do in interviews/algo questions but you should describe what the parameter actually represents so that the code is self-documenting.

frontend/src/components/Review/Rating.tsx
```diff
@@ -0,0 +1,34 @@
...
+  let rows = [];
+  for (let i = 0; i < n; i++) {
+    rows.push(<FavoriteIcon fontSize="small" className={styles.rating} />);
```
ptwu, line 12: For both this and the `empty_hearts` function, you can condense the entire function body to just be `[...Array(n)].map(() =><FavoriteIcon fontSize="small" className={styles.rating} />)`. 

This basically creates an empty array of `n` length and maps over it creating a new copy with `FavoriteIcon` components contained in each element.

ptwu (commented):
Looks good! Just added some code-related comments for you to look at before we merge it into main.

frontend/src/pages/LandlordPage.tsx
```diff
@@ -32,12 +32,8 @@ const LandlordPage = (): ReactElement => {
...
     <div>
       <h1>{`This is dummy text! My current landlordId is ${landlordId}`}</h1>
-      {reviews.map((review) => (
-        <Review
-          overall_rating={review.overall_rating}
-          date={review.date}
-          text={review.text}
-        ></Review>
+      {reviews.map(({ date, text, overall_rating }) => (
+        <Review overall_rating={overall_rating} date={date} text={text} />
```
Enochen, line 36: The prop names are the same, it'd be cleaner to not destructure the argument and use the spread operator instead

frontend/src/components/utils/date.ts
```diff
@@ -0,0 +1,18 @@
+const months = [
+  'Jan',
+  'Feb',
+  'March',
+  'April',
```
Enochen: Generally these are shortened to Mar and Apr if we are taking abbreviations of the other months as well

frontend/src/components/Review/index.tsx
```diff
@@ -2,32 +2,14 @@ import React, { ReactElement } from 'react';
 import styles from './Review.module.scss';
 import Card from 'react-bootstrap/Card';
 import Rating from './Rating';
+import date_to_string from '../utils/date';
```
Enochen: snake_case?? in OUR typescript project?? it's more likely than you think :o

Enochen: overall_rating, filled_hearts, empty_hearts suffer from this as well

frontend/src/pages/LandlordPage.tsx
Enochen: if the argument is (reviewData), then you can do <Review ...reviewData />

ptwu (approved):
LGTM!

PR #31, opened 2021-02-28
Remove reactstrap

Enochen:
### Summary 

For some reason we had both reactstrap and react-bootstrap in our project.

We're probably moving away from both but the reactstrap code is replaced by react-bootstrap equivalents for now

### Notes 

Looks a little different than before (after swapping out components), might be a css issue.


ptwu (approved):
since Bahar said we won't be releasing the FAQ page in our MVP, the different looks should be fine.

PR #32, opened 2021-02-28
[chore] Code cleanup: move images to assets folder; remove default CRA logo

ptwu:
### Summary 

This PR moves all static / image assets to an `assets/` folder in the frontend directory. From now on, we should place all assets here instead of in their respective component folders.

### Test Plan 

Play with staging and see that the functionality and design remains the same.


samliu10 (approved):
Small changes but keeping all static assets/images in one place definitely makes the codebase cleaner, especially since we will be adding more components for the reviews page.

PR #33, opened 2021-02-28
Homepage listings with mobile responsiveness

kjasani111:
## Summary

- Closes issue #14 
- Changes from pull request #18
- This pull request is for the implementation of the homepage design that includes apartment listings, and supports mobile responsiveness.
### Before/After

- The code has been largely refactored to support mobile responsiveness, and is now using the column/grid components provided by the `react-bootstrap` library.
- The design of the listing cards has been changed to match the current design on Figma.
- Currently, the listing cards are populated with hardcoded data, but they have been made functional in a way that populating them with backend data will not be difficult.
### Testing

- Run `yarn install` in `root` to make sure all packages are installed.
- Run `yarn start` in `root` to view the React project in a local browser. It will open to the Home page, where you can see the implementation of the new design.
- Right click the page and select `Inspect` to see how the webpage looks on different sized screens.


ptwu:
The PR is failing the linter and TypeScript compiler, so make sure to fix those errors before we can review it.

frontend/src/App.scss
```diff
@@ -29,3 +41,28 @@ body {
     border-color: #545b62;
   }
 }
+:focus {
+  outline: 0 !important;
```
ptwu, line 45: Usually not a good idea to use `!important` in CSS properties because it usually means you can edit the property in a different part of the UI. We can fix this in a future PR though.

frontend/src/components/Home/ApartmentCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  function renderPrice() {
+    return props.price ? (
+      <Col style={{ paddingRight: '0px', paddingLeft: '10px' }} xs={4}>
+        <Row>
+          <PriceIcon style={{ marginLeft: '12px', marginBottom: '5px' }} />
+          <Card.Subtitle
+            className="ml-1"
+            style={{
```
ptwu, line 40: Try not to use inline styles but put them in a separate CSS file i.e. `ApartmentCard.module.css`.

ptwu (approved):
LGTM! Approving because we should get this merged in before anything else, but left some code-style related comments we can fix in future PRs.

PR #34, opened 2021-03-04
Leave review modal

Enochen:
### Summary 

This diff includes a modal that allows a user to leave a review.

### Test Plan 

image

image

### Notes 

This modal doesn't show up anywhere on the site yet, we can add that when the time comes.

There's no form submitting logic and the modal doesn't know which landlord the review is about (we're not really at that stage yet)

Also, right now the bootstrap css is messing with MUI styles (especially with margins), so once we get rid of bootstrap altogether the grid spacing for this modal will look much less awkward.

~~**(kinda huge)** The modal gets wrecked on a phone screen and I'm not exactly sure how to fix it. The ways of CSS and MUI Grids are beyond me. Hopefully someone can chime in about this~~

~~image~~

^ Fixed, shoutout to @ptwu for knowing how grids work

Today we're going to try three different layouts at three drastically different screen sizes

image

image

image


ptwu:
For the small screens text issue, you can try putting each `<ReviewRating>` in its own grid instead of two in a grid, and giving that grid props of `xs={12}` and `md={6}` so that it takes up all 12/12 of the row on mobile but is half on a medium sized screen.

Referring to the below code for the "two in a grid"
https://github.com/cornell-dti/cu-apts/blob/e736de71e0c7b99856c6e054ee7005557b697c93/frontend/src/components/LeaveReview/ReviewModal.tsx#L118-L129

ptwu:
Also, I'm not sure why the staging GitHub action is failing, since everything compiles and there are no CI errors. Maybe try re-running the action?

Enochen:
> Also, I'm not sure why the staging GitHub action is failing, since everything compiles and there are no CI errors. Maybe try re-running the action?

There's this line: "Treating warnings as errors because process.env.CI = true."

The modal isn't used anywhere, which causes a warning 😒 
I'll try to add a button on the front page that opens this modal, which we can move later.

ptwu:
image

One last thing UI-wise: looks like the content in the card doesn't have `overflow-y: scroll` so a mobile user can't see the review text box input or submit a review.

ptwu (approved):
lgtm!

PR #35, opened 2021-03-04
Set up precommit hook to run Prettier & TSC on all commits

ptwu:
### Summary 

This PR implements a pre-commit hook that runs `yarn format` on every commit, making sure that we don't need to worry about formatting in code review but instead focus on actually important stuff.

All devs should check if this works on their computers (this is an important change for the codebase).

### Test Plan 

Edit some JS/TS/JSX/TSX file and make it badly formatted. Commit stuff as usual and see that the pre-commit hook runs when you commit (before pushing), and fixes the file as per Prettier style guidelines

(looks like this on Windows 10 Git Bash)

image


.husky/pre-commit
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+. "$(dirname "$0")/common.sh"
+
+yarn lint-staged
```
Enochen, line 5: What do you think about also running `yarn tsc` here? This would catch type errors along with whatever prettier does.

Enochen (commented):
Works on my end (Powershell as well as VS Code's git extension)

.husky/pre-commit
ptwu: Good idea, I'll add it

ptwu: this does make commits pretty slow though, so I hope no one is gonna be in a rush

Enochen: Yeah it's pretty slow :/ Lets see how we as the team feel about it after a week or so and then revert it if we don't like it. Definitely depends on how often we commit.

Enochen: By adding `"incremental": true,` to the compiler options of `tsconfig.json`, we can speed up `tsc` runtime by about 2x (for the frontend). It saves a `.tsbuildinfo` file so it doesn't have to do the full thing every time `tsc` gets run.

Once we get rid of bootstrap it'll probably get faster as well.

Enochen (changes requested):
Lets add the tsbuildinfo file to the gitignore

ptwu:
oops forgot to do that. it's in now

Enochen (approved):

Enochen (approved):

PR #36, opened 2021-03-06
Material-UI Navbar

menaattia:
### Summary 

This pull request implements the NavBar for the Home and FAQ pages in Material-UI

- [x] implemented and styled a responsive Material-UI Navbar for the Home and FAQ pages
- [x] deleted the previous implementations NavBar and FAQNavBar

### Test Plan 
Run `yarn start` in the workspace or frontend folders. The NavBar can be seen at `/` and `/faq` routes

homepage2
homepage1
faqpage

### Notes 

When I opened the `/faq` page from the workspace with the backend running, I noticed that the arrow buttons weren't styled like they used to, and I'm not sure what caused this. I didn't make any changes to the CollapsibleQuestion component.
faqarrow


frontend/src/components/utils/NavBar/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+  IconButton,
+  Drawer,
+  Link,
+  MenuItem,
+} from '@material-ui/core';
+import MenuIcon from '@material-ui/icons/Menu';
+import { Link as RouterLink } from 'react-router-dom';
+import icon from '../../../assets/home-icon.png';
+
+export type navbarButton = {
```
ptwu, line 17: nit: best practice for types is to capitalize each first letter of every word in the type name

frontend/src/components/utils/NavBar/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+    marginBottom: '8px',
+  },
+  menuDrawer: {
+    alignSelf: 'right',
+    marginBottom: '8px',
+  },
+}));
+
+const NavBar = ({ headersData }: Props): ReactElement => {
+  const [state, setState] = useState({
```
ptwu, line 71: You should parametrize this state hook using `useState<TypeName>({...})` so that we can't just pass any random type and set the state to that (might have to make a new type with `mobileView: boolean` and `drawerOpen: boolean`)

frontend/src/components/utils/NavBar/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+    menuButton,
+    toolbar,
+    drawerContainer,
+    homeImage,
+    menuDrawer,
+  } = useStyles();
+
+  useEffect(() => {
+    const setResponsiveness = () => {
+      return window.innerWidth < 992
```
ptwu, line 89: We should eventually hoist this up to the utils function and call that function to get this value instead of always doing this computation each time.

frontend/src/pages/HomePage.module.scss
```diff
@@ -1,12 +1,17 @@
...
   background-size: cover;
+  background-color: #f6f6f6;
 }
 
 .search {
   margin-top: 50px;
 }
+
+.homepageDescription {
+  padding-left: 7%;
```
ptwu, line 15: We generally want to shy away from magic percentage numbers like these because they might behave weirdly on mobile. Instead, maybe consider using `px` or `em`.

ptwu (commented):
Left some comments about code style and best practices, but great work!

ptwu:
Thanks for the code style changes! One last thing: I noticed that on both mobile and desktop, the horizontal viewport is getting overflowed (you can see this from the scrollbar at the bottom in the screenshot below). Chances are, one of your CSS properties is using a value of more than `100%` width.

image

ptwu (approved):
lgtm! Thanks for the great work! :)

PR #39, opened 2021-03-12
Info/properties/features component

samliu10:
### Summary 

This pull request includes the 3-section component `InfoFeatures` on the landlord page displaying information about a property/landlord, other properties owned by that landlord, and features of that specific property, e.g. parking, maintenance, laundry, etc. 
- The component uses the `material-ui`Grid layout for responsive sizing. For screen sizes greater than or equal to the `sm` breakpoint, `InfoFeatures` has a column width of 4 and appears to the right of the reviews. At the `xs` breakpoint, `InfoFeatures` takes up the full screen width and moves on top of the reviews. 
- Within `InfoFeatures` are three individual components: `Info` (information about a property/landlord), and two instances of `PropertyInfo`, one showing properties owned by that landlord and the other showing features of the specific property.
    - Horizontal lines separate the three components. 
    - The bullet points within the three components also resize responsively, displaying as either two columns or one column depending on the width of the component. 
  

https://user-images.githubusercontent.com/57203639/110962118-96f79e80-831e-11eb-9705-8dc851cec10d.mov

### Test Plan 
Run `yarn start` in the root directory to start a local dev server. Change the size of the window to see how the components resize.


ptwu:
The CI is failing for staging because there are still linter warnings, and we have strict mode enabled (will probably fix this in a future PR). Could you fix the warning?

PR #40, opened 2021-03-13
Info/properties/features component

samliu10:
### Summary 

This pull request includes the 3-section component `InfoFeatures` on the landlord page displaying information about a property/landlord, other properties owned by that landlord, and features of that specific property, e.g. parking, maintenance, laundry, etc. 
- The component uses the `material-ui`Grid layout for responsive sizing. For screen sizes greater than or equal to the `sm` breakpoint, `InfoFeatures` has a column width of 4 and appears to the right of the reviews. At the `xs` breakpoint, `InfoFeatures` takes up the full screen width and moves on top of the reviews. 
- Within `InfoFeatures` are three individual components: `Info` (information about a property/landlord), and two instances of `PropertyInfo`, one showing properties owned by that landlord and the other showing features of the specific property.
    - Horizontal lines separate the three components. 
    - The bullet points within the three components also resize responsively, displaying as either two columns or one column depending on the width of the component. 
  

https://user-images.githubusercontent.com/57203639/110962118-96f79e80-831e-11eb-9705-8dc851cec10d.mov

### Test Plan 
Run `yarn start` in the root directory to start a local dev server. Change the size of the window to see how the components resize.


frontend/src/pages/LandlordPage.tsx
```diff
@@ -28,13 +31,63 @@ const reviews = [
 
 const LandlordPage = (): ReactElement => {
   const { landlordId } = useParams<Record<string, string | undefined>>();
+  const features = ['Parking', 'Heating', 'Trash removal', 'Snow plowing', 'Maintenance'];
+  const info = ['111 Dryden Rd', '151 Dryden Rd', '418 Eddy St'];
+  const phone = '555-555-5555';
+  const address = '119 S Cayuga St, Ithaca, NY 14850';
+  const [width, setWidth] = useState(window.innerWidth);
```
ptwu, line 38: We already have an `isMobile` util file [here](https://github.com/cornell-dti/cu-apts/blob/main/frontend/src/utils/isMobile.tsx) on the main branch. Maybe you can add a `getWidth` function here that returns the window width so we don't have to create new states every time.

(also if you could, also make both files in the `utils` folder end in `.ts` since they're both not React components, guess I forgot to review that part last time in Mena's PR)

frontend/src/pages/LandlordPage.tsx
```diff
@@ -28,13 +31,63 @@ const reviews = [
 
 const LandlordPage = (): ReactElement => {
   const { landlordId } = useParams<Record<string, string | undefined>>();
+  const features = ['Parking', 'Heating', 'Trash removal', 'Snow plowing', 'Maintenance'];
```
ptwu, line 34: Instead of having separate consts for each individual set of data in a landlord page, maybe make a type for this landlord data and use a `useState<LandlordData>` hook with these default values instead. This will make it easier when we're getting data from the DB.

ptwu (commented):
the implementation looks great! just left two comments about some refactoring to make further dev work easier

ptwu (approved):

PR #41, opened 2021-03-14
Add favicon and document titles helper function hook

ptwu:
### Summary 

This PR implements a default favicon based on our Figma designs as well as a hook allowing for dynamic document titles based on what page the user is on. Simply use `useTitle()` with a string argument or a function determining what it should be.

### Test Plan 

Go to any page.

image

image


samliu10 (dismissed):
Nice job- definitely helpful to have dynamic page titles. It's also good how you made it so `titleOrFn` can be a string or a function returning a string because that makes it more flexible.

samliu10 (approved):
LGTM

PR #42, opened 2021-03-15
Add Unit Tests

menaattia:
### Summary 

This pull request is the first step towards implementing backend unit testing

- [x] installed ts-jest
- [x] updated firestore rules to allow reading but not writing to faq collection
- [x] initialized firestore emulator
- [x] wrote unit tests to check firestore rules
- [x] wrote unit test to check that the faq collection isn't empty 
- [x] wrote unit test to check the status of the get request to `/` route

### Test Plan 
Run `yarn test` from root or backend workspace. There should be 4 tests that pass.
test


package.json
```diff
@@ -47,5 +54,6 @@
     "**/*.{ts,js,tsx,jsx,scss,css,html}": [
       "prettier --write"
     ]
-  }
+  },
+  "dependencies": {}
```
ptwu, line 58: Don't think we need this change

firestore.rules
```diff
@@ -0,0 +1,12 @@
+rules_version = '2';
+service cloud.firestore {
+  match /databases/{database}/documents {
+    match /{document=**} {
+      allow read, write: if false;
+    }
+    match /faqs/{docId} {
+      allow read: if true;
```
ptwu, line 8: These aren't good firestore rules, so maybe make a GitHub issue so we know to fix these rules later on.

ptwu (commented):
Thanks so much for setting the tests up! Just left two comments. Also, to fix the merge conflict, you'll need to run `yarn` in the root directory to solve the `yarn.lock` conflict

ptwu (approved):

ptwu (approved):

PR #43, opened 2021-03-15
add noEmit flag to pre-commit tsc call

Enochen:
### Summary 

Running just `tsc` emits `.js` files in the backend, which messes up with linting and such.

By adding the `--noEmit` flag to the `tsc` command in our precommit hook, we can avoid creating such files.

### Test Plan 

No `.js` files were creating while committing this commit.

### Notes 

`tsc` runs pretty slow still, do we want it in our precommit hook at all?

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -5,4 +5,4 @@
 echo "Prettifying your code on commit..."
 yarn lint-staged
 echo "Running the TypeScript compiler..."
-yarn tsc
+yarn tsc --noEmit


ptwu (approved):
ok looks good, but let's discuss the `yarn tsc` in precommit hook with the team on wednesday. personally, i'm content with CI failing on PRs because it prevents merge into main altogether, but prettier is important in precommit because it can still psas the linter even if prettier isn't run and silently get merged in

PR #47, opened 2021-03-21
Add DB types based on schema

ptwu:
### Summary 

After discussing the schema today in this week's sync meeting, we decided on these types. I've added them to a `db-types.ts` file within a new common directory to be shared between the frontend and backend. This way, when you assume that there's some backend endpoint you can call, you can just do `await response.json() as Review` for example to cast it instantly into a review if you know that it follows a certain structure. Similarly, on the frontend we can now expect that these endpoints return a certain type.

### Test Plan 

n/a

### DIFF
--- a/common/types/db-types.ts
+++ b/common/types/db-types.ts
@@ -1,0 +1,39 @@
+export type DetailedRating = {
+  readonly location: number;
+  readonly safety: number;
+  readonly value: number;
+  readonly maintenance: number;
+  readonly communication: number;
+  readonly conditions: number;
+};
+
+export type Review = {
+  readonly id: string;
+  readonly aptId: string | null;
+  readonly landlordId: string;
+  readonly overallRating: number;
+  readonly detailedRatings: DetailedRating;
+  readonly reviewText: string;
+  readonly date: Date;
+};
+
+export type Landlord = {
+  readonly id: string;
+  readonly name: string;
+  readonly contact: string | null;
+  readonly avgRating: number;
+  readonly photos: readonly string[]; // can be empty
+  readonly reviews: readonly string[]; // array of Review IDs in reviews collection
+  readonly properties: readonly string[]; // array of Apartment IDs in apartments collection
+};
+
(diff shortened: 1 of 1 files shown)


common/types/db-types.ts
```diff
@@ -0,0 +1,39 @@
...
+};
+
+export type Apartment = {
+  readonly id: string;
+  readonly name: string;
+  readonly address: string; // may change to placeID for Google Maps integration
+  readonly landlordId: string | null;
+  readonly numBaths: number | null;
+  readonly numBeds: number | null;
+  readonly photos: string[]; // can be empty
```
Enochen, line 37: should this be readonly?

Enochen (changes requested):

common/types/db-types.ts
ptwu: yeah, the idea is that when you fetch stuff from the DB you shouldn't be mutating it anyway since it won't actually cause any changes to the data. If we need to alter any fields, we can just use spreading syntax to create a new immutable copy.

Enochen: Oh I meant `readonly photos: readonly string[]`

common/types/db-types.ts
```diff
@@ -0,0 +1,39 @@
...
+  readonly detailedRatings: DetailedRating;
+  readonly reviewText: string;
+  readonly date: Date;
+};
+
+export type Landlord = {
+  readonly id: string;
+  readonly name: string;
+  readonly contact: string | null;
+  readonly avgRating: number;
```
menaattia, line 24: we might want to add numRatings to use it to calculate the avgRating

common/types/db-types.ts
Enochen: If we don't add the second readonly we will still be able to mutate it which isn't what we want

ptwu: oops i forgot to add it there, will do it now

Enochen (approved):
LG™

PR #48, opened 2021-03-24
add carousel and btns

Enochen:
### Summary 

This diff adds an image carousel that appears when a user clicks a button labeled "show all photos" (the position of this button is just randomly placed as a placeholder, as we don't have a header yet to put it into).

It also adds the aforementioned button as well as a button to open up the "leave a review" modal on the Landlord page.

### Test Plan 

image

image

image

image


frontend/src/components/Review/InfoFeatures.tsx
```diff
@@ -19,16 +18,14 @@ export default function InfoFeatures({
...
-          <PropertyInfo title="Property Info" info={propertyInfo} />
-          <div className={styles.divider}></div>
-          <PropertyInfo title="Property Features" info={propertyFeatures} />
-        </CardContent>
-      </Card>
-    </Grid>
+    <Card variant="outlined">
+      <CardContent>
+        <Info phone={phone} address={address} />
+        <div className={styles.divider}></div>
```
ptwu, line 24: this is an element with no contents, so you can just do `<div ... />`

frontend/src/pages/LandlordPage.tsx
```diff
@@ -73,70 +78,87 @@ const dummyRatingInfo: RatingInfo[] = [
 
 const LandlordPage = (): ReactElement => {
   const { landlordId } = useParams<Record<string, string | undefined>>();
-  const [width, setWidth] = useState(window.innerWidth);
   const [landlordData] = useState(dummyData);
   const [aveRatingInfo] = useState(dummyRatingInfo);
-  const breakpoint = 600;
+
+  const [reviewOpen, setReviewOpen] = useState(false);
```
ptwu, line 84: Might want to parametrize these `useState` hooks so that they don't get inferred to just be a falsy value.

ptwu (commented):
Looks great! Can you resolve the merge conflict so we can see the staging site? Also left a few nits below.

frontend/src/pages/LandlordPage.tsx
Enochen: Which hooks are you talking about? landlordData and aveRatingInfo?

frontend/src/components/Review/InfoFeatures.tsx
Enochen: ah this is not code that I wrote, I think MUI already has a <Divider /> so I'll use that instead

frontend/src/pages/LandlordPage.tsx
ptwu: Lines 84-85: you can do `const [reviewOpen, setReviewOpen] = useState<boolean>(false);` for example

Enochen: Is this needed? Typescript seems to infer the type if you put "false"
image

ptwu: ok my bad, it only does this when you supply an empty array as the init value. disregard this

ptwu (approved):

PR #49, opened 2021-03-28
Leave review POST request

samliu10:
### Summary 

This pull request is for the POST request that submits a review to the backend. The review must be of type `Review` specified in `db-types.ts`. 

### Test Plan 
Run `yarn start` in `backend` to start a local dev server. Test the endpoint using a POST request in Postman. The Firebase database will update with the new review in real-time.

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -11,18 +11,18 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "@firebase/testing": "^0.20.11",
     "@types/cors": "^2.8.8",
     "@types/express": "^4.17.8",
+    "@types/supertest": "^2.0.10",
     "body-parser": "^1.19.0",
     "cors": "^2.8.5",
     "dotenv": "^8.2.0",
     "express": "^4.17.1",
     "firebase": "^8.0.1",
     "firebase-admin": "^9.3.0",
+    "supertest": "^6.1.3",
     "ts-node": "^9.1.1",
-    "typescript": "^4.0.5",
-    "@firebase/testing": "^0.20.11",
-    "@types/supertest": "^2.0.10",
-    "supertest": "^6.1.3"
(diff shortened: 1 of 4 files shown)


backend/src/app.ts
```diff
@@ -1,12 +1,13 @@
 import express, { Express } from 'express';
-import bodyParser from 'body-parser';
 import cors from 'cors';
 import { db } from './firebase';
 import { Section } from './firebase/types';
+import { Review } from '../../common/types/db-types';
 
 const app: Express = express();
+const reviews = db.collection('reviews');
```
ptwu, line 8: make this more indicative that it's a collection (`reviews` is kind of generic)

ptwu (commented):
looks good, just one naming nit

backend/package.json
```diff
@@ -11,18 +11,18 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "@firebase/testing": "^0.20.11",
```
Enochen, line 14: According to [the npm page](https://www.npmjs.com/package/@firebase/testing), this package is deprecated

backend/src/app.ts
```diff
@@ -29,4 +30,11 @@ app.get('/', async (req, res) => {
   res.status(200).send(JSON.stringify(faqs));
 });
 
+app.post('/new-review', async (req, res) => {
+  const doc = reviews.doc();
+  const review: Review = req.body;
```
Enochen, line 35: Not sure if we're at this stage yet but we probably should destructure the properties we want. This prevents us from storing extra/non-schema-related properties that attackers might include in the request body.

ptwu: alternatively, you can just do `const review = req.body as Review`

ptwu (approved):

PR #50, opened 2021-03-28
Get Reviews from Backend

menaattia:
### Summary 

This pull request is the first step towards getting reviews from the backend

- [x] factored out the code to make get requests in a get function in the utils folder
- [x] used the get function in FAQPage to get FAQs
- [x] changed the types of props to match the Review and DetailedRating types in db-types

### Test Plan 

The function currently cannot be tested on getting reviews because there is no backend for getting reviews yet, but it is also used for getting FAQs: `yarn start` in the root folder and the FAQs from the backend should appear


frontend/src/utils/get.ts
```diff
@@ -0,0 +1,12 @@
+import axios from 'axios';
+
+export default function get<T>(route: string, setState: (data: T[]) => void) {
+  axios
+    .get(`http://localhost:8080${route}`)
```
ptwu, line 5: Maybe make the backend URL a constant in this file, so we don't have to update it for all of the HTTP methods

frontend/src/pages/LandlordPage.tsx
```diff
@@ -115,11 +126,14 @@ const LandlordPage = (): ReactElement => {
   const { landlordId } = useParams<Record<string, string | undefined>>();
   const [landlordData] = useState(dummyData);
   const [aveRatingInfo] = useState(dummyRatingInfo);
-
+  const [reviewData, setReviewData] = useState<Review[]>(reviews);
```
ptwu, line 129: We don't relaly have a need for the constant / hardcoded data for reviews anymore because of the work done in this PR. Instead, the initial value of this state can just be `[]`

ptwu (approved):
looks good!

PR #51, opened 2021-03-29
Header for landlord page

kjasani111:
### Summary
This pull request is for the header for the landlord review page. I have added a basic header that has photos, the landlord name, the landlord reviews, and a button to view more photos. It is implemented in Material UI and supports mobile, and the data props are set up in such a way so that they can accept dynamic data from the backend.

### Test Plan
Run `yarn start` from root to see the website. Click on the reviews button to see the review header on the reviews page.


frontend/src/components/Landlord/Header.tsx
```diff
@@ -0,0 +1,132 @@
+import React, { ReactElement } from 'react';
+import HeartRating from '../utils/HeartRating';
+import ApartmentPhoto1 from '../../assets/apartment_exterior.png';
```
Enochen, line 3: Since we're going to store links to the landlord photos in our database, let's replicate that by adding/using dummy image links from landlordData instead of adding the dummy images to the project

Enochen:
image

The circle becomes an oval when you squish the aspect ratio to that of mobile devices

This shouldn't happen if you use the [Avatar component](https://material-ui.com/components/avatars/) which makes these circles for you

frontend/src/pages/LandlordPage.tsx
```diff
@@ -8,12 +8,16 @@ import Review from '../components/Review/Review';
...
+import LandlordHeader from '../components/Landlord/Header';
 
 type LandlordData = {
   properties: string[];
   photos: string[];
   phone: string;
   address: string;
+  name: string;
+  overallRating: number;
+  numReviews: number;
```
menaattia, line 20: There's a Landlord type in common/types/db-types, which includes the name and overallRating fields and has a reviews field which is an array you can get the length of instead of adding the field numReviews, so maybe use that Landlord type instead of LandlordData

frontend/src/components/Landlord/Header.tsx
```diff
@@ -0,0 +1,132 @@
...
+  landlord_rating: {
+    '& div': {
+      display: 'inline-block',
+      marginTop: '2px',
+    },
+  },
+});
+
+const LandlordHeader = ({ landlord, overallRating, numReviews }: Props): ReactElement => {
+  const classes = useStyles();
```
menaattia, line 87: maybe deconstruct classes `let { media, logo, photoButton, ... etc } = useStyles();` so you can refer to the classes directly

frontend/src/components/Landlord/Header.tsx
```diff
@@ -0,0 +1,132 @@
...
+    paddingLeft: 0,
+    paddingBottom: 0,
+    paddingTop: '20px',
+    fontStyle: 'normal',
+    fontWeight: 'bold',
+    fontSize: '36px',
+    lineHeight: '43px',
+    letterSpacing: '0.02em',
+  },
+  landlord_reviews: {
```
menaattia, line 68: This and landlord_rating on line 78 should also be camelCase

menaattia:
In addition to Enoch's comment about the circle being oval, for some widths, the hearts leave the container and overlap with other components
Screen Shot 2021-03-30 at 1 46 55 AM

frontend/src/components/Landlord/Header.tsx
```diff
@@ -0,0 +1,132 @@
...
+      marginTop: '2px',
+    },
+  },
+});
+
+const LandlordHeader = ({ landlord, overallRating, numReviews }: Props): ReactElement => {
+  const classes = useStyles();
+  return (
+    <Grid container spacing={0}>
+      <React.Fragment>
```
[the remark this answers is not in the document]
kjasani111: Yes, it is necessary for the GlobalCSS component declared just below that. That's how MaterialUI said to do it in their guide.

frontend/src/components/Landlord/Header.tsx
```diff
@@ -0,0 +1,132 @@
...
+    border: '2px solid black',
+    borderColor: 'black',
+    boxSizing: 'border-box',
+    borderRadius: '8px',
+    padding: '0px',
+    '&:hover': {
+      background: 'rgba(255, 255, 255, 0.8)',
+    },
+    '&:focus': {
+      borderColor: 'black !important',
```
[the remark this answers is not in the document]
kjasani111: The reason I had to use !important here is because the button default for focus in MaterialUI has an !important borderColor that gets set to gray, so to override that, I had to use !important as well. Let me know if you have any other suggestions, or if that is ok.

kjasani111:
For now I have resolved all of the frontend issues because Peter said I can ignore the nit issues for now, but we can do the backend changes to the props in a future PR

ptwu (approved):
Looks sensible for now. Consider doing a clean code PR later that fixes more code quality nits pointed out above, but great work!

samliu10 (dismissed):
Yeah that plan sounds good! Approving this

ptwu (approved):

PR #54, opened 2021-04-04
Backend endpoint for querying by landlord or apartment

menaattia:
### Summary 

This pull request is the first step towards implementing autocomplete search on the homepage

- [x] implemented `/reviews/` backend route to query by landlord or apartment using fuse.js
- [x] implemented a `/landlords` post method

### Test Plan 

Run the backend server by running `yarn start` in the backend folder. Make GET requests to `localhost:8080/reviews/?q=query` in Postman. This should return all the landlords and apartments that contain the query. For testing purposes, I put into the database landlords with names Joseph, James, Joshua, John. Querying `j` returns all 4. Querying `jo` returns Joseph, John and Joshua, and `jos` returns Joseph and Joshua. I also added Collegetown Terrace to the buildings collection, and Carole to lanlords. Querying `c` returns both Collegetown Terrace and Carole. The queries are case insensitive; you could also query with uppercase letters. (Because fuse.js is fuzzy search, other landlords/apts are also returned that may not match the exact query).

To test the POST request, post to `/landlords` in Postman with a body of Landlord type. This returns the id of the created doc and in the firebase console, you can see that the landlord gets added to the landlords collection.


backend/src/server.test.ts
```diff
@@ -1,3 +1,4 @@
+// eslint-disable-next-line
```
Enochen, line 1: There doesn't seem to be any linter errors when this is removed

backend/src/app.ts
```diff
@@ -48,4 +51,31 @@ app.get('/reviews/:idType/:id', async (req, res) => {
   res.status(200).send(JSON.stringify(reviews));
 });
 
+app.post('/landlords', async (req, res) => {
```
Enochen, line 54: It is probably a good idea to try/catch all of these handlers, so that it won't crash the function or return a 201 when something goes wrong

backend/src/app.ts
```diff
@@ -48,4 +51,31 @@ app.get('/reviews/:idType/:id', async (req, res) => {
...
+  const doc = landlordCollection.doc();
+  const landlord: Landlord = req.body as Landlord;
+  const searchableKeywords: string[] = generateKeywords(landlord.name);
+  const data = { ...landlord, searchableKeywords };
+  doc.set(data);
+  res.status(201).send(doc.id);
+});
+
+app.get('/reviews/:query', async (req, res) => {
```
ptwu, line 63: We should use query params instead of route params since it's more idiomatic for this use-case (`/reviews/` and then access through `req.query.something`)

backend/src/utils/generateKeywords.ts
```diff
@@ -0,0 +1,17 @@
+export default function generateKeywords(name: string): string[] {
```
ptwu, line 1: while this is clever, and I think this is a cool way to assist in searchability, I think we should just defer to a fuzzy-search library such as Fuse.js (https://fusejs.io/) to handle keyword matching

Enochen: Don't libraries like this require the entire piece of data (all apts/landlords) loaded into memory?

ptwu: That's true. Honestly, since the number of landlords in Ithaca won't be big enough for it to not be loadable into memory, we can just handle it in the backend since we don't need scalability to justify Algolia. (also this is just an MVP)

backend/src/app.ts
```diff
@@ -48,4 +51,35 @@ app.get('/reviews/:idType/:id', async (req, res) => {
...
+    const doc = landlordCollection.doc();
+    const landlord: Landlord = req.body as Landlord;
+    doc.set(landlord);
+    res.status(201).send(doc.id);
+  } catch (err) {
+    res.status(400).send(err);
+  }
+});
+
+app.get('/reviews/', async (req, res) => {
```
ptwu, line 65: Remove the trailing `/` at the end of the URI since we just need to call `reviews?q=something`

ptwu (approved):

PR #55, opened 2021-04-11
Add common font theme and fix DOM nesting errors + other code quality improvements

ptwu:
### Summary 

This pull request adds Work Sans as our common font theme in accordance with the Figma designs. I've imported font weights 300, 400, and 600 for use. I've also modified the navbar to better match the one in the MVP designs, though without the SVG icon yet (it's been silently added to the assets folder, however)

I've also fixed that pesky DOM nesting error in the console by removing the containers within the `Typography` components in the apartment cards. This was silently causing the `<p>` and `<div>` nesting error. In addition, I've removed all inline styles, since this is bad practice.

Mobile responsiveness was also silently failing at least in Chrome, so I fixed that with a meta viewport tag.

### Test Plan 

Rb8kG2dEUp


samliu10 (approved):
Nice job- font definitely makes the website look a lot better, and it's good that you caught the issue with mobile responsiveness failing. Still have more things to fix up with the navbar to fully align it with the designs, e.g. svg, active links, the slogan under the logo, but that's separate from this. Everything here looks good to me

PR #56, opened 2021-04-14
Img upload

Enochen:
### Summary 

This diff implements image upload for reviews.

The leave review modal has been updated with an upload files button.

Uploaded files are uploaded on the client side to Cloud Storage. Only authenticated users may upload photos.

On the backend, the review schema has been updated with a string list `photos` of photo urls 

### Test Plan 

Photos are can be uploaded and viewed:
image
image
image

### Notes 

For now, only the first photo of a review is displayed. This will be updated when there is a design for the case when there are multiple photos.

This diff also implements a small fix on the backend for converting the date in request bodies from `string` to `Date` so that we store a Firebase timestamp instead of a string.


ptwu (commented):
Looks good and functionality is working on my end. One thing is, the user doesn't know what image file types are supported: I tried uploading a `.webp` image and it ended up just setting the `photos` array to be empty. Maybe put some kind of indication of what file types are accepted (JPG, PNG, GIF) on the frontend UI and then check this on the backend?

samliu10 (commented):
Nice job 👍

One thing though: It's ok if we don't have this for the mvp, but one thing that would be nice to have in the future is an option to remove an uploaded picture when a user is writing the review. Say you uploaded a  few files but then before you submit the review, you decide you don't want to post all of them. Right now the only way to change uploaded files is to close the modal and start over.

Enochen:
> Nice job 👍
> 
> One thing though: It's ok if we don't have this for the mvp, but one thing that would be nice to have in the future is an option to remove an uploaded picture when a user is writing the review. Say you uploaded a few files but then before you submit the review, you decide you don't want to post all of them. Right now the only way to change uploaded files is to close the modal and start over.

If you click "Choose Files" and then exit out of the upload file prompt, the list of files should be cleared. This IS pretty unintuitive tho, and we should definitely revisit this to take care of the user experience.

Enochen:
> Looks good and functionality is working on my end. One thing is, the user doesn't know what image file types are supported: I tried uploading a `.webp` image and it ended up just setting the `photos` array to be empty. Maybe put some kind of indication of what file types are accepted (JPG, PNG, GIF) on the frontend UI and then check this on the backend?

I finally tested this and I was able to get a `.webp` image uploaded. Can you try to replicate and send me the image that doesn't get uploaded? I suspect that it is a file that is too large; if that's the case we should be able to bring up a toast or error message if an upload fails (and not submit the review)

ptwu:
> > Looks good and functionality is working on my end. One thing is, the user doesn't know what image file types are supported: I tried uploading a `.webp` image and it ended up just setting the `photos` array to be empty. Maybe put some kind of indication of what file types are accepted (JPG, PNG, GIF) on the frontend UI and then check this on the backend?
> 
> I finally tested this and I was able to get a `.webp` image uploaded. Can you try to replicate and send me the image that doesn't get uploaded? I suspect that it is a file that is too large; if that's the case we should be able to bring up a toast or error message if an upload fails (and not submit the review)



> > Looks good and functionality is working on my end. One thing is, the user doesn't know what image file types are supported: I tried uploading a `.webp` image and it ended up just setting the `photos` array to be empty. Maybe put some kind of indication of what file types are accepted (JPG, PNG, GIF) on the frontend UI and then check this on the backend?
> 
> I finally tested this and I was able to get a `.webp` image uploaded. Can you try to replicate and send me the image that doesn't get uploaded? I suspect that it is a file that is too large; if that's the case we should be able to bring up a toast or error message if an upload fails (and not submit the review)

Seems to work now. That's weird. Maybe the image itself wasn't uploaded due to human error on my part in that review. Let's merge this in - it looks good and from my testing seems to be pretty intuitive UI-wise. One thing to discuss on Sunday is whether we really want users to be able to upload 3 photos (this PR only supports 1 photo)

ptwu (approved):

PR #59, opened 2021-04-18
Batch script for adding new reviews to db

samliu10:
### Summary 

This pull request is for a batch script that adds new reviews to the database. `script.ts` loads a JSON file of reviews from the google drive spreadsheet `'landlord_reviews.json'` that is located in the same directory `src`. This script formats each entry into a review of type `Review` with `id` being the auto-generated id from Firebase and posts it to the database. 

### Test Plan 

Run `yarn script` in `backend` to run `script.ts` on a local dev server. Then check the database to see the new reviews.  
 **Important: running this command will add all reviews in the JSON to the database, so do not run this unless you intend to do that.** 

Close the server and run `yarn start` in the root directory to see the new reviews on the website. This video shows new reviews for the landlord with `landlordId` 24. 

https://user-images.githubusercontent.com/57203639/115158112-9b6c5100-a05a-11eb-9731-8076bbb6e4c9.mov

### Notes 

This script assumes that there are no null entries in the JSON file and that the spreadsheet contains entries with the following schema: 
```
{
   id: "",
  aptId: number,
  landlordId: number,
  overallRating: number,
  date: string,
  detailedRatings.location: number,
  detailedRatings.safety: number,
  detailedRatings.value: number,
  detailedRatings.maintenance: number,
  detailedRatings.communication: number,
  detailedRatings.conditions: number,
  reviewText: string
}
```


backend/src/script.ts
```diff
@@ -0,0 +1,54 @@
+import express, { Express } from 'express';
+import cors from 'cors';
+import { db } from './firebase-config';
+import reviewData from './landlord_reviews.json';
+import { Review } from '../../common/types/db-types';
+
+const reviewCollection = db.collection('reviews');
+
+const app: Express = express();
```
ptwu, line 9: There's no need to initialize an Express server for the script, since we don't need to call any endpoints or have any need for middleware

backend/src/script.ts
```diff
@@ -0,0 +1,54 @@
...
+    maintenance: data['detailedRatings.maintenance'],
+    communication: data['detailedRatings.communication'],
+    conditions: data['detailedRatings.condition'],
+  },
+  date: data.date,
+});
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const modifyData = async (data: any[]) => {
+  const reviews = data.map((review) => formatReview(review));
```
ptwu, line 45: you can do this all in one pass if you just call `makeReview` on the result of each `formatReview`

ptwu (approved):
lgtm!

PR #60, opened 2021-04-18
Add API request logging to Express server

ptwu:
### Summary 

This PR adds HTTP request logging to our backend, as a quality of life feature so that you can tell when we're polling too much in a `useEffect` hook or something in the future, or if needless calls are being made in general.

### Test Plan 

Run `yarn start` to run the frontend and backend concurrently. Try doing something in the frontend that fetches from the Express server and see that it gets logged as expected.

image

### Notes

Make sure to run `yarn install` after this PR gets merged and you pull from `main`!

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -4,7 +4,7 @@
   "description": "",
   "main": "server.ts",
   "scripts": {
-    "tsc": "tsc", 
+    "tsc": "tsc",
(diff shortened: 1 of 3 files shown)


samliu10 (approved):
Oo this is super nice to have. Now we can see all of the requests being made, and the logs are really detailed. :)!

PR #61, opened 2021-04-30
Redirect to landlord review page after searching

menaattia:
### Summary 

- [x] created new Apartment and Landlord types with a type field that is either `'LANDLORD '`or `'APARTMENT'`
- [x] used the new types to add labels to the options listed in the autocomplete search
- [x] implemented get request to get an `ApartmentWithId` from an aptId
- [x] implemented logic to redirect to landlord page if the selected option is a landlord, and get the landlord id associated with an apt if the selected option is an apt, and redirect to the landlord associated with it 

### Test Plan 
Run `yarn start` in the root folder and enter queries into the search bar. Select an option by clicking on it or hitting enter and whether you select a landlord or an apt, it redirects to a landlord review page. Some possible queries: landlord, collegetown.

https://user-images.githubusercontent.com/31854132/116662316-9f09ad00-a963-11eb-9137-5410ef4caa03.mov

### Notes

- The redirect only works when clicking on a menu item or hitting enter, but it does not work by clicking the search icon. I'm not quite sure what we want to do with the search icon. If there are no matches to the query and they hit search, what do we want to happen? I was thinking that as the user goes down the menu with the up and down arrows, the option that is being focused should appear in the search bar, and redirect when they click the search icon, I tried implementing this with an onFocus on the menuItem setQuery to that option, but the up and down arrows stopped working, so I would need to figure out how to implement that with focus. 
- When the selected option is an apartment, even though it has a landlordId field, it cannot be accessed because the option has a type of `LandlordWithType | ApartmentWithType` and landlord doesn't have a landlordId field. This is why I set up a get request to get an ApartmentWithId from an aptId and access its landlordId field.

### DIFF
--- a/backend/src/app.ts
+++ b/backend/src/app.ts
@@ -4,7 +4,14 @@ import Fuse from 'fuse.js';
 import morgan from 'morgan';
 import { db } from './firebase-config';
 import { Section } from './firebase-config/types';
-import { Review, Landlord, Apartment } from '../../common/types/db-types';
+import {
+  Review,
+  Landlord,
+  LandlordWithId,
+  LandlordWithType,
+  ApartmentWithType,
+  ApartmentWithId,
+} from '../../common/types/db-types';
 import authenticate from './auth';
 
 const app: Express = express();
@@ -58,6 +65,17 @@ app.get('/reviews/:idType/:id', async (req, res) => {
   res.status(200).send(JSON.stringify(reviews));
 });
 
+app.get('/apts/:id', async (req, res) => {
+  try {
+    const { id } = req.params;
+    const snapshot = await aptCollection.doc(id).get();
+    const aptDoc = { id, ...snapshot.data() } as ApartmentWithId;
+    res.status(200).send(JSON.stringify(aptDoc));
+  } catch (err) {
+    res.status(400).send(err);
+  }
+});
+
 app.post('/landlords', async (req, res) => {
   try {
     const doc = landlordCollection.doc();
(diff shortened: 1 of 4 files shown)


frontend/src/components/Home/Autocomplete.tsx
```diff
@@ -28,17 +37,20 @@ export default function Autocomplete() {
...
   const [width, setWidth] = useState(inputRef.current.offsetWidth);
+  const [selectedId, setSelectedId] = useState('');
 
-  const handleClose = (event: React.MouseEvent<EventTarget>, option: Apartment | Landlord) => {
-    const element = event.currentTarget as HTMLInputElement;
-    setQuery(element.innerText);
-    setSelected(option);
-    setOpen(false);
-  };
+  // const handleClose = (
```
ptwu, line 46: this commented code is unneeded now, so you can remove it

backend/src/app.ts
```diff
@@ -69,21 +87,39 @@ app.post('/landlords', async (req, res) => {
   }
 });
 
+const isLandlord = (obj: LandlordWithId | ApartmentWithId): boolean => {
+  const keys = Object.keys(obj);
+  return keys.includes('contact');
```
ptwu, line 92: This function is `O(n)` which is unnecessary for just checking whether the object has a `contact` field or not. You can instead just do it in constant time using `if ('contact' in obj) { }`

backend/src/app.ts
```diff
@@ -69,21 +87,39 @@ app.post('/landlords', async (req, res) => {
...
+};
+
 app.get('/reviews', async (req, res) => {
   try {
     const query = req.query.q as string;
     const landlordDocs = (await landlordCollection.get()).docs;
-    const landlords: Landlord[] = landlordDocs.map((landlord) => landlord.data() as Landlord);
+    // eslint-disable-next-line
```
ptwu, line 99: you can easily fix these eslint errors by removing the curly brackets. If you omit the `return` in an arrow function in JS, and your function is just a value, JS assumes that you're returning that value on each iteration. For this example, it'd be

```
 const landlords: LandlordWithId[] = landlordDocs.map((landlord) => (
      { id: landlord.id, ...landlord.data() } as LandlordWithId
 ));
```

common/types/db-types.ts
```diff
@@ -2,6 +2,10 @@ type Id = {
   readonly id: string;
 };
 
+type Type = {
```
ptwu, line 5: Are we thinking of adding more inner types to this parent `Type`? Not sure why it's a single type nested inside another. Also, `Type` is a very vague name; maybe consider something more specific

ptwu (commented):
Looks great functionality-wise!! Really impressed by how great it looks, even on mobile. I've added some code quality comments for you to address, as well as one performance change.

ptwu (approved):

PR #64, opened 2021-05-02
Helpful button (likes system)

Enochen:
### Summary 

This diff lets users mark reviews as "helpful"

The database structure behind this is a `likes` counter for each review and a collection of documents for each user containing the reviews they have liked. The backend runs a Firestore transaction to make sure that the `likes` counter is synchronized with the user likes document. This prevents likes from being duplicated (from spamming server requests) due to interweaving.

### Test Plan 
(not sure why my cursor is so off in the gif)
image

### Notes 

Obviously new things have been added to the database schema.
There shouldn't be any breaking changes.


backend/src/app.ts
```diff
@@ -115,7 +122,54 @@ app.get('/reviews', async (req, res) => {
...
+        likesRef.set({ [reviewId]: FieldValue.delete() }, { merge: true });
+        reviewRef.update({ likes: FieldValue.increment(-1) });
+      }
+    });
+    res.status(200).send(JSON.stringify({ result: 'Success' }));
+  } catch (err) {
+    console.error(err);
+    res.status(400).send('Error');
   }
 });
```
ptwu, line 174: consider putting the body of the functions here in a helper function with maybe a boolean on which operation you're doing in order to deduplicate the code

frontend/src/components/Landlord/Header.tsx
```diff
@@ -124,13 +124,7 @@ const LandlordHeader = ({ name, overallRating, numReviews, handleClick }: Props)
...
                 }}
               >
-                <Grid
-                  container
-                  xs={12}
-                  direction="row"
-                  style={{ height: '200px' }}
-                  alignItems="flex-end"
-                >
+                <Grid container direction="row" style={{ height: '200px' }} alignItems="flex-end">
```
ptwu, line 127: nit: avoid inline styles

frontend/src/pages/LandlordPage.tsx
```diff
@@ -98,6 +102,48 @@ const LandlordPage = (): ReactElement => {
...
+          review.id === reviewId ? { ...review, likes: (review.likes || 1) - 1 } : review
+        )
+      );
+      const token = await user.getIdToken(true);
+      setLikedReviews((reviews) => ({ ...reviews, [reviewId]: false }));
+      axios.post('/remove-like', { reviewId }, createAuthHeaders(token));
+    } catch (err) {
+      console.log('error with liking review');
+    }
+  };
```
ptwu, line 145: a helper here could also be useful to prevent duplicated code

ptwu (commented):
smart implementation, and looks good! left some code quality comments.

frontend/src/utils/get.ts
```diff
@@ -1,14 +1,12 @@
 import axios from 'axios';
 
-export const backendUrl = 'http://localhost:8080';
-
 export default function get<T>(
   route: string,
-  setState: (data: T[]) => void,
-  setLoading: ((loading: boolean) => void) | undefined
+  setState: (data: T) => void,
+  setLoading: ((loading: boolean) => void) | undefined = undefined
```
samliu10, line 6: Oh nice this is better. I noticed that most of the times we were using `get`, we were always passing in `undefined` anyways. Definitely will make my `landlord-cards` pr cleaner- there are so many GET requests in there.

ptwu (approved):

PR #65, opened 2021-05-02
Landlord cards

samliu10:
### Summary 

This pull request populates the three apartment cards on the Home page with building data from the db. The cards show the name of the property, address, renting company, and the number of reviews associated with the company. Each card links to Landlord page showing all reviews for that company.  Since there are only three cards, data for three random buildings and their associated landlords is used to populate the cards. This implementation assumes there are at least three buildings in the db. 

### Test Plan 
- Run `yarn start` in the root directory to start a local dev server. Click on a card to see the reviews associated with that renting company. 

https://user-images.githubusercontent.com/57203639/116830362-87a30d80-ab77-11eb-96dd-4b9fff14cec5.mov

### Note 
- The endpoint to get reviews by id `'/reviews/:idType/:ids'` now has route param `ids` which is a string list of ids separated by commas. This allows us to retrieve reviews for multiple buildings/landlords at once.

### DIFF
--- a/backend/src/app.ts
+++ b/backend/src/app.ts
@@ -43,7 +43,7 @@ app.post('/new-review', authenticate, async (req, res) => {
     doc.set({ ...review, date: new Date(review.date) });
     res.status(201).send(doc.id);
   } catch (err) {
-    res.status(401).send('Error');
+    res.status(400).send('Error');
   }
 });
 
@@ -51,9 +51,8 @@ app.get('/reviews/:idType/:id', async (req, res) => {
   const { idType, id } = req.params;
   const reviewDocs = (await reviewCollection.where(`${idType}`, '==', id).get()).docs;
   const reviews: Review[] = reviewDocs.map((doc) => {
-    let data = doc.data();
-    data = { ...data, date: data.date.toDate() };
-    return data as Review;
+    const { date, ...data } = doc.data();
+    return { date: date.toDate(), ...data } as Review;
   });
   res.status(200).send(JSON.stringify(reviews));
 });
--- a/backend/src/app.ts
+++ b/backend/src/app.ts
@@ -4,7 +4,7 @@ import Fuse from 'fuse.js';
 import morgan from 'morgan';
 import { db } from './firebase-config';
 import { Section } from './firebase-config/types';
-import { Review, Landlord, Apartment } from '../../common/types/db-types';
+import { Review, Landlord, Building } from '../../common/types/db-types';
 import authenticate from './auth';
 
 const app: Express = express();
@@ -75,8 +75,8 @@ app.get('/reviews', async (req, res) => {
     const landlordDocs = (await landlordCollection.get()).docs;
     const landlords: Landlord[] = landlordDocs.map((landlord) => landlord.data() as Landlord);
     const aptDocs = (await aptCollection.get()).docs;
-    const apts: Apartment[] = aptDocs.map((apt) => apt.data() as Apartment);
-    const aptsLandlords: (Landlord | Apartment)[] = [...landlords, ...apts];
+    const apts: Building[] = aptDocs.map((apt) => apt.data() as Building);
+    const aptsLandlords: (Landlord | Building)[] = [...landlords, ...apts];
 
     const options = {
       keys: ['name', 'address'],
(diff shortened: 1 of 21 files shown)


frontend/src/pages/HomePage.tsx
```diff
@@ -1,9 +1,83 @@
...
 import Autocomplete from '../components/Home/Autocomplete';
+import { Apartment, LandlordWithId, Review } from '../../../common/types/db-types';
+import { Link as RouterLink } from 'react-router-dom';
+import get from '../utils/get';
 
 const HomePage = (): ReactElement => {
+  const [homeData, setHomedata] = useState<any>([]);
+  const [buildingData, setBuildingData] = useState<Apartment[]>([]);
+  const [landlordData, setLandlordData] = useState<LandlordWithId[]>([]);
+  const [reviewsBuilding1, setReviewsBuilding1] = useState<Review[]>([]);
```
ptwu, line 13: We probably don't need a state for each of these buildings: you can simply make one type of `useState<[Review, Review, Review]>` which is basically a tuple in TypeScript

frontend/src/pages/HomePage.tsx
```diff
@@ -21,21 +95,37 @@ const HomePage = (): ReactElement => {
...
+              let company = '';
+              let numReviews = 0;
+              if (landlordId) {
+                company = idToLandlord[landlordId].company;
+                numReviews = allReviews[index].length;
+              }
+              return (
+                <Grid item xs={12} md={4}>
+                  <Link
+                    {...{
```
ptwu, line 109: since none of these properties are dynamic, there's no need to spread out the object. does it not work if you just do the same as props for the `Link` component?

frontend/src/pages/HomePage.tsx
```diff
@@ -21,21 +95,37 @@ const HomePage = (): ReactElement => {
...
-          </Grid>
-          <Grid item xs={12} md={4}>
-            <ApartmentCard address="117 Eddy St" bedsAndBaths="1 Br | 2 B" />
-          </Grid>
-          <Grid item xs={12} md={4}>
-            <ApartmentCard address="117 Eddy St" bedsAndBaths="5 Br | 2 B" price="$800" />
-          </Grid>
+          {loaded &&
+            buildingData.map(({ name, address, landlordId, numBaths, numBeds, photos }, index) => {
+              let company = '';
```
ptwu, line 100: nit: you can make these `const`s if you just use a ternary operator. for example: `const company = landlordId ? idToLandlord[landlordId].company : ''`

backend/src/app.ts
```diff
@@ -119,4 +120,27 @@ app.get('/reviews', async (req, res) => {
   }
 });
 
+app.get('/homepageData', async (req, res) => {
+  const buildingDocs = (await aptCollection.limit(3).get()).docs;
+  const buildings: Apartment[] = buildingDocs.map((doc) => doc.data() as Apartment);
+  // eslint-disable-next-line consistent-return
```
ptwu, line 126: these are a lot of linter disables. i know that you want to do things in one-pass (so filtering and then mapping wouldn't be ideal), but you can still do it in one-pass without having these linter disables using `reduce` - try accumulating from an empty array and pushing to it when the condition is met

frontend/src/pages/HomePage.tsx
```diff
@@ -1,9 +1,83 @@
...
+  useEffect(() => {
+    if (buildingData && buildingData.length > 0 && allReviews) {
+      setLoaded(true);
+    }
+  }, [buildingData, allReviews]);
+
+  let idToLandlord: { [id: string]: { company: string; reviews: readonly string[] } } = {};
+
+  landlordData &&
+    // eslint-disable-next-line array-callback-return
```
ptwu, line 74: same here

frontend/src/pages/HomePage.tsx
```diff
@@ -1,9 +1,83 @@
...
+
+  useEffect(() => {
+    if (buildingData && buildingData.length > 2) {
+      get<Review>(
+        `/reviews/landlordId/${buildingData[2].landlordId}`,
+        setReviewsBuilding3,
+        undefined
+      );
+    }
+  }, [buildingData]);
```
ptwu, line 59: lots of repeated code here that can be deduplicated by just making the building states a single state as described above

ptwu (commented):
looks great! left some code quality nits / fixes that can help us make it more maintainable

ptwu (approved):

PR #66, opened 2021-05-02
Fix UI/console bugs, decrease search bar label length

ptwu:
### Summary 

This PR fixes a frontend console error that was logging because of the `xs` property being improperly applied to non Grid Item elements, as well as a search bar bug where if you press the down arrow key, it scrolls down the page instead of going down the list of autocomplete results. In addition, it decreases the length of the search bar label so that it doesn't overflow on mobile small screens.

### Test Plan 

Run `yarn start`. Check that the console doesn't log any errors anymore when visiting a landlord review page. Press the down arrow key when searching and it should not scroll the entire page down while going through the autocomplete results. Also, simulate a mobile view and see that the search text works.

### DIFF
--- a/frontend/src/components/Home/Autocomplete.tsx
+++ b/frontend/src/components/Home/Autocomplete.tsx
@@ -44,8 +44,8 @@ export default function Autocomplete() {
   const [selectedId, setSelectedId] = useState('');
 
   function handleListKeyDown(event: React.KeyboardEvent) {
+    event.preventDefault();
     if (event.key === 'Tab') {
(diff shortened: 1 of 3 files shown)


samliu10 (approved):
I went through and tested everything- seems good to me. One thing maybe we could do in the future for better UI for the search bar is to bring the search bar forward a little more by adding a shadow?? Not sure though. It's just that right now the results show up on top of the apartment cards, and it looks very flat. But definitely something for the future!

Screen Shot 2021-05-02 at 10 57 20 PM

Issue #63, opened 2021-05-02
[frontend] No input validation on review submit


ptwu:
Currently, it's possible to create a review with empty fields and nothing filled out. We should require at least the overall experience rating to be filled out, as well as review text.

my empty review:
![image](https://user-images.githubusercontent.com/7517829/116825170-0e49f180-ab5c-11eb-8acf-d3d113c46490.png)

PR #67, opened 2021-05-05
Handle ITU reviews with no ratings, implement read more/less

menaattia:
### Summary 

- [x] implemented read more/ read less for reviews above 500 characters (arbitrary number could be changed)
- [x] set minimum height for review content to 200px (can also be changed) to make short reviews seen
- [x] handled the display of reviews with no ratings to only display the review text
- [x] Added a label to reviews with no ratings stating they are ITU reviews
- [x] Removed the icon button from the search bar and moved the search icon to the start of the textfield

### Test Plan 
Run `yarn start` in root folder and go to `/landlord/1`. There should be 2 reviews with no ratings: a short one and a long one. The long review has a Read More button that expands the text and read less that shortens expanded text.

Screen Shot 2021-05-05 at 3 16 41 AM


ptwu (commented):
Code quality looks great! Just one design nit: the icon on the LHS is a bit too close to the text and doesn't have a lot of space from the text. Will approve once this is fixed.

design:
image

implementation
image

frontend/src/components/Review/Review.tsx
```diff
@@ -55,50 +57,73 @@ const ReviewComponent = ({ review }: Props): ReactElement => {
...
       <Card className={root} variant="outlined">
-        <CardContent>
-          <Grid container spacing={2}>
-            <Grid item container justify="space-between">
-              <Grid container spacing={2}>
+        <Box minHeight="200px">
+          <CardContent>
+            <Grid container spacing={2}>
+              <Grid item container justify="space-between">
+                {overallRating !== null ? (
```
samliu10, line 64: There's one review in the db that technically has all ratings of 0, not null, so it's still showing up. I highlighted it here in the console. But this probably isn't something you have to change in this PR because I think it's more of an issue of users being able to submit unrated reviews right now, and that's mentioned in [issue #63 ](https://github.com/cornell-dti/cu-apts/issues/63)

Screen Shot 2021-05-06 at 4 42 17 PM

frontend/src/components/Review/Review.tsx
```diff
@@ -55,50 +57,73 @@ const ReviewComponent = ({ review }: Props): ReactElement => {
...
+                    </CardContent>
+                  </Collapse>
                 </Grid>
+              )}
+              <Grid item container alignContent="center">
+                <Typography>
+                  {expandedText ? reviewText : reviewText.substring(0, 500)}
+                  {!expandedText && reviewText.length > 500 && '...'}
+                  {reviewText.length > 500 ? (
+                    <Button className={button} onClick={() => setExpandedText(!expandedText)}>
```
samliu10, line 106: Nice job with this! It's good that the button has hover effects. One thing we probably should do in the future for better UI is to fade the text behind the "Read More" button so that it's more obvious what the functionality is. 
Screen Shot 2021-05-06 at 4 50 54 PM

kjasani111:
One more design nit about the dropdown arrow. I was testing it on mobile, and for the Galaxy Fold the arrow moves down to the next line. 
Screen Shot 2021-05-08 at 10 10 53 PM
Also for the dropdown arrow, when it is clicked it opens downward instead of to the side like in the design, and the icon isn't the same
Design
Screen Shot 2021-05-08 at 10 13 14 PM
Implementation
Screen Shot 2021-05-08 at 10 12 58 PM

frontend/src/components/Review/Review.tsx
```diff
@@ -55,50 +57,73 @@ const ReviewComponent = ({ review }: Props): ReactElement => {
...
+              {photos.length > 0 && (
+                <Grid container alignItems="center" justify="center">
+                  <Grid item xs={12} sm={6}>
+                    <CardMedia
+                      component="img"
+                      alt="Apt image"
+                      image={photos[0]}
+                      title="Apt image"
+                    />
+                  </Grid>
```
kjasani111, line 121: Just one more nit here. Currently we are only displaying the first photo for a review, but in the design there is a carousel for multiple photos:
Screen Shot 2021-05-08 at 10 29 22 PM
 I don't think we should change this for this PR, but maybe we can include this in a future PR. Great code quality though!

kjasani111:
One more design nit about the dropdown arrow. I was testing it on mobile, and for the Galaxy Fold the arrow moves down to the next line. 
Screen Shot 2021-05-08 at 10 10 53 PM
Also for the dropdown arrow, when it is clicked it opens downward instead of to the side like in the design, and the icon isn't the same.
Design
Screen Shot 2021-05-08 at 10 13 14 PM
Implementation
Screen Shot 2021-05-08 at 10 12 58 PM

frontend/src/components/Review/Review.tsx
menaattia: Yeah, null ratings are for ITU reviews. We shouldn't be allowing users to post ratings of 0, like you said.

ptwu (approved):

PR #68, opened 2021-05-05
Update homepage to current design

ptwu:
### Summary 

This PR implements the latest designs as per Figma. It updates all the homepage components to better match the design, and also does some chores to make the code quality better (inline styles, etc). 

### Test Plan 

scr

👀 

### Notes

I've omitted some of the metadata associated with apartment cards (e.g. address, num baths and beds) since these are no longer part of the design, and also would be hard to source without a lot of manual data pruning for individual apts in a building.


samliu10 (approved):
Nice job with this! It looks so much better now. And good idea to take out the number of beds and baths- it didn't really make sense to have that because the cards represent buildings, not specific units. Next step is to add the icon and slogan 💯

ptwu:
> Nice job with this! It looks so much better now. And good idea to take out the number of beds and baths- it didn't really make sense to have that because the cards represent buildings, not specific units. Next step is to add the icon and slogan 💯 

yep, Kinjal's working on it and should be done soon. tysm!

PR #72, opened 2021-05-16
Enforce non-empty reviews

menaattia:
### Summary 

This pull request closes issue #63 

- [x] Added a check in the backend to not allow posting reviews with no overall rating or review text

### Test Plan 
Run `yarn start` in root folder and go to `/landlord/1`. Try posting review with no overall rating but with review text, no overall rating and no review text, overall rating and no review text. The form shouldn't be submitted in all three cases. The form gets successfully submitted otherwise.

### DIFF
--- a/backend/src/app.ts
+++ b/backend/src/app.ts
@@ -52,6 +52,9 @@ app.post('/new-review', authenticate, async (req, res) => {
   try {
     const doc = reviewCollection.doc();
     const review = req.body as Review;
+    if (review.overallRating === 0 || review.reviewText === '') {
+      res.status(401).send('Error: missing fields');
+    }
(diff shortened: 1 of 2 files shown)


ptwu (changes requested):
Looks good from the backend side, but it'll be weird for any users on the client-side since there are no frontend indications that the inputs are validated. We need to add some `error` prop to the corresponding `TextField` if the input is bad on submit.

samliu10 (commented):
Yeah adding onto what Peter said, the user will see the red error toast that indicates there's an error, but it doesn't say what the error is. Definitely should have error messages above the fields to tell them what's wrong.

ptwu (approved):

PR #73, opened 2021-05-16
Batch script for adding new landlords to db

samliu10:
### Summary 

This pull request is for a batch script that adds new landlords to the database.`script.ts` loads a JSON file of landlords from  `landlords.json` that is located in the directory `./data`.  In the JSON file, each object has an id that represents the id of the landlord. This script formats each entry into a landlord of type `Landlord` and posts it to the database, setting the id of the new document to be the id of that landlord from the JSON object.

Note: the google drive spreadsheet of landlords was converted from csv to a JSON file to get `landlords.json`. That's why landlords without photos, reviews, or properties become JSON objects with these fields set to the empty string and so must be converted into empty arrays. 

Screen Shot 2021-05-16 at 4 53 52 PM

### Test Plan 
Comment out line 68 in `script.ts` to avoid accidentally posting new reviews to the database. Then run `yarn script` in `backend` to run `script.ts` on a local dev server. Then check the database to see the new landlords.  
 **Important: running this command will add all landlords in the JSON to the database, so do not run this unless you intend to do that.**


ptwu (approved):

backend/src/script.ts
```diff
@@ -32,4 +34,37 @@ const formatReview = (data: any): Review => ({
...
+const formatLandlord = ({
+  id,
+  name,
+  contact,
+  avgRating,
+  photos,
+  reviews,
+  properties,
+}: // eslint-disable-next-line @typescript-eslint/no-explicit-any
+any): LandlordWithId => ({
```
Enochen, line 46: It might be helpful to make a type representing the json objects

backend/src/script.ts
```diff
@@ -32,4 +34,37 @@ const formatReview = (data: any): Review => ({
...
+  id: id.toString(),
+  name,
+  contact,
+  avgRating: avgRating.length > 0 ? Number.parseFloat(avgRating) : 0,
```
Enochen, line 50: This can be `avgRating: Number.parseFloat(avgRating || 0)`

Same for the rest, for example `photos: photos || []`

backend/src/script.ts
```diff
@@ -32,4 +34,37 @@ const formatReview = (data: any): Review => ({
...
+  reviews: reviews.length > 0 ? reviews : [],
+  properties: properties.length > 0 ? properties : [],
+});
+
+const makeLandlord = async (landlordWithId: LandlordWithId) => {
+  try {
+    const { id, ...rest } = landlordWithId;
+    const doc = landlordCollection.doc(id);
+    const landlord = rest as Landlord;
+    doc.set({ ...landlord });
```
Enochen, line 61: `.set({ ...landlord })` is just `set(landlord)`

backend/src/script.ts
```diff
@@ -32,4 +34,37 @@ const formatReview = (data: any): Review => ({
...
+  } catch (err) {
+    // eslint-disable-next-line no-console
```
Enochen, line 63: The eslint config for `/backend` has `no-console` turned off so there is no reason to disable it anywhere.

backend/src/script.ts
```diff
@@ -32,4 +34,37 @@ const formatReview = (data: any): Review => ({
...
+  photos,
+  reviews,
+  properties,
+}: // eslint-disable-next-line @typescript-eslint/no-explicit-any
+any): LandlordWithId => ({
+  id: id.toString(),
+  name,
+  contact,
+  avgRating: avgRating.length > 0 ? Number.parseFloat(avgRating) : 0,
+  photos: photos.length > 0 ? photos : [],
```
Enochen, line 51: If `photos` is a non-empty string, you would be writing a string instead of a list into the db, which would mess things up. This applies to the other properties in the json with a string as the value.

PR #92, opened 2021-10-28
Fix autocomplete search result back button

samliu10:
### Summary 

This PR makes a fix to the autocomplete search bar on the home page. The previous implementation uses `Redirect` which does not preserve the history. When a user clicks on a search result, navigated to the landlord page, and clicked the back button, they are not taken back to the home page but to some page in the history. See video below.

https://user-images.githubusercontent.com/57203639/139335480-e02d822f-3873-4320-9965-83bd73a64436.mov

This implementation uses `Router-Link` from `react-router-dom` to handle page history. Now, the user is taken back to the home page. See video below. 

https://user-images.githubusercontent.com/57203639/139335563-2f4bd846-50b6-4139-9813-9ce2c222ee66.mov

Also includes minor CSS edits: building addresses now appear in gray, and the search menu dropdown has a borderShadow. These are not necessary, so let me know if you prefer how it was before!

### DIFF
--- a/frontend/src/components/Home/Autocomplete.tsx
+++ b/frontend/src/components/Home/Autocomplete.tsx
@@ -1,5 +1,4 @@
 import React, { useEffect, useState, useRef } from 'react';
-import { Redirect } from 'react-router-dom';
 import {
   Chip,
   CircularProgress,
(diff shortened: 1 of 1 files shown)


kjasani111 (approved):
This looks good to me! I also do like the CSS changes, as they make it easier to view the addresses.

menaattia (approved):
Lgtm!

PR #93, opened 2021-10-31
Sorting Reviews

menaattia:
### Summary 

This pull request is the first step towards implementing sorting options for reviews

- [x] This pull request closes issue #
- [x] implemented a sorting function that takes in a review object and a property and sorts the reviews based on that provided property
- [x] implemented a Dropdown component for the different options for sorting reviews (most recent and most helpful for now)
- [x] made the review components get re-rendered each time a sorting option is chosen

### Test Plan 

Run `yarn start` in the root folder. Choose renting companies from the homepage. On the renting company review page, click on the "sort reviews by" button and select each of the two options. 
- `landlord/24` has no liked reviews and initially the reviews aren't sorted by date. After clicking sort by most recent, they should become sorted. Sorting by most helpful changes the order of the reviews arbitrarily (unstable sorting) when there's ties.
- `landlord/`10` has one liked review and it should move to the top when sorting by most helpful
- `landlord/16` has one review so sorting doesn't change anything
- `landlord/30` has no reviews so clicking on the sorting options does nothing

### Notes 

[Resolved]
- ~~There is a bug that I found only when sorting by most helpful. Steps to replicate:~~
- ~~Sort `landlord/10` by most helpful. The most liked review moves to the top. Sort by most helpful again and it becomes unsorted.~~
- ~~I'm not sure why this is happening because the onClick for sorting by most helpful only calls the sort function and sets the reviewData to the value returned by the sort function, and the sort function calls Javascript's sort function~~


frontend/src/pages/LandlordPage.tsx
```diff
@@ -69,6 +70,18 @@ const LandlordPage = (): ReactElement => {
...
 
+  useEffect(() => {}, [reviewData]);
+
+  type Fields = keyof typeof reviewData[0];
+
+  const sortReviews = (arr: ReviewWithId[], property: Fields) => {
+    let unsorted = arr;
+    return unsorted.sort((r1, r2) => {
+      // @ts-ignore: Object possibly null or undefined
+      return r1?.[property] < r2?.[property] ? 1 : -1;
```
samliu10, line 81: I think the reason why it's getting sorted out of order is because `likes` doesn't exist on every `Review` object: check out the picture I attached in the comment, specifically the highlighted part. 
For each `(r1, r2)` being compared, I logged `r1?.id + r2?.id + r1?.[property] + r1?.[property] + (r1?.[property] < r2?.[property] ? 1 : -1)` to see how they are being sorted.  `w5d0lTDUmsai7xkw1kAG` is the id of the review that has `likes: 1`. When compared to a review without property `likes`, e.g. `v42nUJUW9pWFcaI0SQkA` where `r1?.[likes]` evaluates to `undefined`, it gets sorted behind the review with the undefined value when it should be put in front.

samliu10: For the "Sort Reviews By" section, what do you think about making it like this where the text that's displayed is the sorting criteria being used? We could have it default to sort by most recent and either have "Sort By" be static or don't include that.  I'm fine with either option since we don't have a designer right now. 
Option 1: 
Screen Shot 2021-10-30 at 9 53 24 PM
 
Option 2:
IMG_0393

frontend/src/pages/LandlordPage.tsx
```diff
@@ -161,6 +174,23 @@ const LandlordPage = (): ReactElement => {
           >
             Leave a Review
           </Button>
+          <DropDown
+            label="sort reviews by"
```
samliu10, line 178: For the "Sort Reviews By" section, what do you think about making it like this where the text that's displayed is the sorting criteria being used? We could have it default to sort by most recent and either have "Sort By" be static or don't include that.  I'm fine with either option since we don't have a designer right now. 
Option 1: 
Screen Shot 2021-10-30 at 9 53 24 PM
 
Option 2:
IMG_0393

kjasani111 (changes requested):
Also, in terms of the style, can you change it so that the button has a different color so it's easier to see how big it is (maybe gray)? Also can you add an arrow next to the "sort reviews by" text? I feel like that would make it similar to the sort buttons on other sites. Also, can we move it closer to the box for the first review? I feel like it looks a little strange right below the "leave a review" button.

samliu10 (approved):
LGTM! Good feature to have for release.

kjasani111 (approved):
It looks good to me! This will be a convenient feature for users trying to navigate the site at release.

Issue #84, opened 2021-10-17
Add profanity filter


samliu10:
We want to prevent users from submitting reviews containing profanity. 

AC: 
- Do not allow a user to submit a review that contains profanity. Use [`bad-words`](https://www.npmjs.com/package/bad-words) to detect if a review contains profanity, and if it does, show an error and do not allow the form to submit. 
- CU Reviews uses this package. See some of their references in [forms](https://github.com/cornell-dti/course-reviews-react-2.0/blob/master/server/endpoints/Review.ts) and [exports](https://github.com/cornell-dti/course-reviews-react-2.0/blob/master/common/profanity.js)

PR #94, opened 2021-10-31
Profanity filter

kjasani111:
### Summary 

This pull request implements checking for profanity before submitting a review.

- [x] Closes issue #84 
- [x] Implemented a profanity filter that checks for profanity in the onSubmit() function in the ReviewModal. 
- [x] Created a type for the profanity filter that can be modified to add additional words to the filter later on.
- [X] The current modal from the main branch allowed the user to submit an empty review if they submit a valid review and click the leave a review and the submit button again without entering review text or an overall experience rating. This bug was fixed in this PR.

### Test Plan 
Run `yarn start` in the root folder. Choose renting companies from the homepage. On the renting company review page, click on the "leave a review" button. 
- Try pressing the "submit" button and confirm that it does not allow submission. 
- Try selecting a number for the overall experience, try submitting again, and confirm it doesn't allow submission, and the helper text below the review text field says "This field is required"
- Try writing a profane word in the review text, and try submitting. Confirm that submission is not allowed, and the helper text says "This review contains profanity. Please edit it and try again."
- Remove the profane word, and write something normal. Try submitting, and confirm the submission works.
- Press the "leave a review" button again, and try pressing the "submit" button without selecting an overall experience rating, and without typing anything in the review field. Confirm that the submission is not allowed.

### DIFF
--- a/frontend/src/utils/profanity.ts
+++ b/frontend/src/utils/profanity.ts
@@ -1,0 +1,5 @@
+const Filter = require('bad-words');
+
+const filter = new Filter();
+
+export const includesProfanity = (s: string) => filter.isProfane(s);
--- a/frontend/src/components/LeaveReview/ReviewModal.tsx
+++ b/frontend/src/components/LeaveReview/ReviewModal.tsx
@@ -167,6 +167,7 @@ const ReviewModal = ({ open, onClose, setOpen, landlordId, onSuccess, toastTime 
       }
       console.log(review);
       setOpen(false);
+      dispatch({ type: 'reset' });
       onSuccess();
     } catch (err) {
       console.log(err);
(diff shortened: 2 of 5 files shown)


samliu10 (dismissed):
Great work!! It's important for launch + post-launch that the reviews we get are clean, and this is a good way for us to validate the content we are serving.

frontend/src/utils/profanity.ts
```diff
@@ -0,0 +1,5 @@
+const Filter = require('bad-words');
```
menaattia, line 1: I think here you can put `import Filter from 'bad-words';` instead to make it consistent with the other imports.

menaattia (commented):
Looks good! Just left two minor code quality comments.

menaattia (dismissed):

PR #98, opened 2021-11-03
Terms and conditions: privacy policy page

samliu10:
### Summary 

This PR adds the privacy policy page we need in order to release this product. It is consistent with Course Plan's description on their website here: [https://courseplan.io/policy](https://courseplan.io/policy). 

- Adds privacy policy
- Minor CSS change to the footer so that the footer is center aligned on mobile
Screen Shot 2021-11-03 at 9 20 39 AM

### Test Plan 

- Navigate to `/policies` by clicking on "Privacy Policy" in the footer


frontend/src/pages/Policies.tsx
```diff
@@ -0,0 +1,270 @@
...
+  },
+  link: {
+    color: '#B94630',
+  },
+}));
+
+const Policies = (): ReactElement => {
+  const { h2, h3, h4, body, link } = useStyles();
+  return (
+    <Container maxWidth="lg">
```
menaattia, line 50: I think it would be better to add left and right padding (7% maybe?) to this container so the text isn't too close to the edges. Other than that, looks good!

samliu10: I left it like that so the text was aligned with the navbar text- let me know if you still think we should add padding!

menaattia: Oh I see. I don't think it matters that much, but I still think it would look nicer if we had some space on the sides.

samliu10: This is what 7% looks like
Screen Shot 2021-11-07 at 7 39 39 PM

frontend/src/components/utils/Footer/index.tsx
```diff
@@ -16,24 +18,42 @@ const useStyles = makeStyles(() => ({
...
+        </Grid>
+        <Grid item>
+          <Link
+            {...{
+              to: `/policies`,
+              style: { color: 'black' },
+              component: RouterLink,
+            }}
+          >
+            <Typography> Privacy Policy </Typography>
```
kjasani111, line 55: Style Nit: Can this be underlined to distinguish it from the line above (The "Got a question that isn’t answered?" line). I feel like that would make it easier to see.

kjasani111, line 55: This seems to not line up with the rest of the footer for certain screen sizes on mobile, particularly for the iPhone 5/SE and the Galaxy Fold. See the screenshots below:
Screen Shot 2021-11-07 at 11 45 34 PM
Screen Shot 2021-11-07 at 11 45 42 PM
Can we fix this so that it lines up for the smaller screen sizes?

frontend/src/pages/Policies.tsx
```diff
@@ -0,0 +1,270 @@
...
+      <Typography variant="h2" className={h2}>
```
kjasani111, line 51: This is just a suggestion, but would this look better if we centered the title? I'm not sure if the privacy policy needs to follow a certain format though, so it's fine if we can't do this.

kjasani111 (changes requested):

frontend/src/pages/Policies.tsx
kjasani111: I think I like it how it is now actually

frontend/src/pages/Policies.tsx
samliu10: I'll keep it as is to stay consistent with the rest of the website since that is all left-aligned!

frontend/src/components/utils/Footer/index.tsx
samliu10: I can't figure out how to fix it on super small screen sizes e.g. Galaxy Fold. But I got it to work on all other sizes. 
Screen Shot 2021-11-08 at 10 26 43 AM

samliu10: I tried it out, but I think I'll keep it as is because of the hover styling that underlines this link and "contact".

frontend/src/pages/Policies.tsx
samliu10: Alright I'll leave it as it is then.

frontend/src/components/utils/Footer/index.tsx
kjasani111: Yeah that sounds good to me!

kjasani111 (approved):
Everything looks good now!

PR #99, opened 2021-11-03
JSON file for new reviews

samliu10:
### Summary 

This PR includes a new json file of the new reviews we have since the last time reviews were written to the db in May 2021. All reviews in this file have been added to the db according to the instructions in #59


menaattia (approved):
Looks good!

PR #101, opened 2021-11-04
Remove indent on property info 

samliu10:
### Summary 

Makes a minor CSS change by removing the indentation on the side of the property names. 
Before: 

Screen Shot 2021-11-03 at 8 59 15 PM
After: 
Screen Shot 2021-11-03 at 8 57 11 PM


menaattia (approved):
lgtm!

PR #111, opened 2021-11-10
Sign in immediately after pressing "leave a review"

### LINKED ISSUE #110: Sign in immediately after clicking "leave a review"
Currently, users are only prompted to sign in after writing a review and clicking "submit". Users may prefer being asked to sign in first after clicking "leave a review" and opening the modal in case they can't log into their account or something.

samliu10:
### Summary 

This PR closes #110 and changes the sign in flow so that now when a user clicks "Leave A Review", they are immediately requested to sign in instead of after filling out the form in the modal and pressing "Submit". 
- An error pops up if a user tries to log in with a non Cornell email, and the modal does not open. The error tells them to sign in with a Cornell email. 

https://user-images.githubusercontent.com/57203639/141041279-0b356d36-2a8f-4ca3-9953-0d52a6dd4116.mov

### Test Plan 

Go incognito
- You should not be able to sign in with a non Cornell email, and an error should show. 
- If you click on "Leave a Review" after attempting to sign in with a non Cornell email, the pop up should still open, and you should be able to select a new email to sign in with. 
- You should see the error every time you try to sign in with a non Cornell email. 
- Signing in with a Cornell email should redirect to Cornell's SSO and then open the modal. 

### Note
[Line 32](https://github.com/cornell-dti/cu-apts/blob/main/frontend/src/utils/firebase.ts#L32) in function `getUser` signs the user out of firebase if they attempt to log in with a non Cornell email. This is NOT the same as logging out of Google. The user is unable to make any changes to the db, e.g. submit a review, upvote a review because they are not logged into firebase, but they remain logged into their Google account. We don't want to force them to sign out of Google in case they have other tabs open, so the workaround is to have the sign-in popup allow users to select the email they want to sign in with every time they click "Leave a Review" — see image below. Without this, the popup opens and closes immediately because the user is technically already signed in, and they are unable to sign into any other account.
Screen Shot 2021-11-10 at 8 48 12 PM


frontend/src/pages/LandlordPage.tsx
```diff
@@ -131,6 +141,18 @@ const LandlordPage = (): ReactElement => {
 
   const removeLike = likeHelper(true);
 
+  const openReviewModal = async () => {
+    if (!user) {
+      let user = await getUser(true);
+      setUser(user);
+      if (!user) {
```
samliu10, line 148: `getUser` returns the user if it's a Cornell email. Otherwise, returns `null`.

frontend/src/components/LeaveReview/ReviewModal.tsx
```diff
@@ -149,7 +150,6 @@ const ReviewModal = ({ open, onClose, setOpen, landlordId, onSuccess, toastTime
   const onSubmit = async () => {
     try {
       setSending(true);
-      const user = await getUser(true);
       if (!user) {
```
menaattia, line 153: I don't think we need this null check in the onSubmit function anymore since you added it to openReviewModal function and it shows a sign-in error toast when the user is null

menaattia (commented):
Looks good! I tested it by logging into my Cornell email and personal email and everything works as expected. I just left one minor comment.

kjasani111 (dismissed):
All of the code looks good to me! I just had a suggestion about user friendliness. When I was testing and signed in with a non Cornell email, the error toast did show, but it was relatively fast and kind of small at the bottom of the screen, so I missed it the first time. Maybe we can make it bigger and/or show for longer? This is just a suggestion though.

frontend/src/components/LeaveReview/ReviewModal.tsx
samliu10: Good point!

samliu10:
> All of the code looks good to me! I just had a suggestion about user friendliness. When I was testing and signed in with a non Cornell email, the error toast did show, but it was relatively fast and kind of small at the bottom of the screen, so I missed it the first time. Maybe we can make it bigger and/or show for longer? This is just a suggestion though.

Changed it so that the toast shows up at the top of the page and for a little longer!

frontend/src/pages/ApartmentPage.tsx
```diff
@@ -98,13 +100,21 @@ const ApartmentPage = (): ReactElement => {
 
   type Fields = keyof typeof reviewData[0];
 
-  const showConfirmationToast = () => {
-    setShowConfirmation(true);
+  const showToast = (setState: (value: React.SetStateAction<boolean>) => void) => {
```
samliu10, line 103: Ok that this stuff is repeated between `ApartmentPage.tsx` and `LandlordPage.tsx` for now. There are no more common updates between the two pages, and soon `LandlordPage.tsx` will look very different. A lot of it will be changed, and all of the currently repeated stuff will not be applicable to the page anymore.

menaattia (dismissed):
lgtm!

menaattia (dismissed):

menaattia (approved):

PR #112, opened 2021-11-15
Skeleton apartment page

### LINKED ISSUE #102: Set up initial apartment page with reviews
Right now, the website shows reviews aggregated by landlord because we initially had the problem of not having enough reviews for each property. It will be a good idea to change this so that reviews are aggregated by building name instead. We just want to show the reviews for a specific apartment — other fixes, e.g. recording the apt id when leaving a review, will be addressed in another issue. 
- Populate apartment page with reviews for that apartment

samliu10:
### Summary 

This PR closes #102 and sets up a skeleton apartment page `ApartmentPage.tsx` for the foundations of an apartment-based reviews approach that can be accessed from `/apartment/:aptId`. This page contains everything `LandlordPage.tsx` contains EXCEPT
- Reviews are for the specific apartment, not the landlord. They are retrieved using the `aptId` property of type `Review`. 
- `useEffect` calls are updated to retrieve information needed for an apartment-based approach, e.g. apartment data, reviews for that apartment, etc. 
- The website title now shows the name of the apartment, e.g. "Eddygate" 

All other changes that need to be made to this page — info in the page header (gray block in the pic), the info panel on the right side, edits to the review modal — will be addressed in separate issues/PRs. This is so that PRs are kept small. 

In the picture below, only one review shows up because there is only one review with `aptId = 2` which corresponds to Eddygate. 

Screen Shot 2021-11-15 at 5 22 42 PM

### Test Plan 

- Navigate to `/apartment/:aptId` where `:aptId` can be 2, 3, or 4 because those are the only buildings currently in the `buildings` collection. 
- Reviews will not show up for `/apartment/3` or `apartment/4` because there are no reviews with `aptId = 2` or `aptId = 3`. This problem is being addressed separately through database cleaning.


menaattia (approved):
The skeleton looks good and mostly follows the Landlord page, with changes made to adapt it to apartments. I noticed you changed the title to the apartment name and paid attention to other details such as "reviews" vs. "review" for a single review. I made changes to the LandlordPage to fix the bug related to sorting reviews by most helpful and the changes aren't reflected here, so I can make these changes here also in a separate PR (as soon as my PR that addresses the bug is reviewed to make sure it's all good before we make the change here as well), but approving for now.

kjasani111 (approved):
This looks good to me for now! I liked the coding style and using useEffect() because that makes the components easy to update. I also liked that each function has a descriptive name and that each function body is small, which makes it easy to debug for future changes and easy to understand for other developers. I do have one suggestion/possible bug though. If I go to apartment/5 or landlord/50 or some other id that doesn't exist in the database, the page just stays in linear progress forever. I don't know if that is good behavior. Should we load the 404 page instead if the id isn't found? This doesn't need to be addressed for this PR but maybe we can look into it in future PRs.

samliu10:
> This looks good to me for now! I liked the coding style and using useEffect() because that makes the components easy to update. I also liked that each function has a descriptive name and that each function body is small, which makes it easy to debug for future changes and easy to understand for other developers. I do have one suggestion/possible bug though. If I go to apartment/5 or landlord/50 or some other id that doesn't exist in the database, the page just stays in linear progress forever. I don't know if that is good behavior. Should we load the 404 page instead if the id isn't found? This doesn't need to be addressed for this PR but maybe we can look into it in future PRs.

This is a really good suggestion! I'll make an issue for this.

PR #116, opened 2021-11-17
Home page data optimization

### LINKED ISSUE #113: Optimize data retrieval for apartment cards
Right now, the apartment cards on the home page take a long time to load on first load. Look into optimizing data retrieval for the buildings in the db. 

Also should look into optimizing search. It's much slower now because there are more buildings in the db.

samliu10:
### Summary 

This PR closes #113 by making an optimization to the data retrieval for the route `page-data/:page` that is currently being used on the home page. A request to this route on the home page returns data on three buildings in the `buildings` collection, and a request to this route on the "reviews" page returns data on all buildings. 

- The previous implementation required `n` db queries to the `reviews` collection, where `n` is the number of buildings to be retrieved, i.e. for each building, we made a request to `reviews` collection. 
- Now only one db query is required to retrieve all reviews associated with the buildings, i.e. we can get the necessary reviews in one pass. 
  - Note: this optimization assumes that `# reviews >> # buildings`. 

### Test Plan 

- See if the apartment cards on the home page load faster. Check specifically on the firebase staging link since usually that's really slow. 
- We will have a better idea of speed once we add all buildings to the db. Right now there are only three.


thuypham03:
Great improvement! Decrease the number of query significantly help reduce loading time of the website.

huyl3:
LGTM

PR #122, opened 2021-11-28
New reviews and landlord/buildings script fix

samliu10:
### Summary 

This PR makes the following changes 

- Adds the new review JSON data that contains the correct values for `aptId`. This data has been written to the db. 
- Makes a formatting change to the landlord and building scripts. Without this change, each document is written as an object that looks like: 
```
{ 
  building: 
  {
      id: number;
      name: string;
      landlordId: number;
      address: string;
      area: string;
      photos: string[];
      numBeds: number;
      numBaths: number;
  } 
}
```
when it should not have the outer JSON around it, i.e. should just look like 
```
{
    id: number;
    name: string;
    landlordId: number;
    address: string;
    area: string;
    photos: string[];
    numBeds: number;
    numBaths: number;
} 
```


menaattia (approved):
Looks good, but there are a few linter warnings for unused variables: `photos`, `numBeds`, `numBaths` that we might want to get rid of

menaattia (approved):

PR #126, opened 2021-11-30
Temporary building limit

samliu10:
### Summary 

This PR adds a temporary limit on the number of buildings and their data fetched from the `buildings` collection in the DB. This limit is necessary to prevent excessively high amounts of reads until we optimize the way we retrieve this data. See #116.


menaattia (approved):
lgtm

kjasani111 (approved):
Looks good!

PR #127, opened 2021-12-02
Apartment cards fix: number of reviews, routes 

### LINKED ISSUE #125: Change info on apartment cards to show reviews for apartment
Apartment cards currently show the number of reviews for the landlord of the apartment, not for the apartment itself. This will require changing the route `/page-data/:page`.

samliu10:
### Summary 

This PR closes #125 and makes the following changes for apartment cards: 
- Changes route `page-data/:page` to retrieve the number of reviews for the apartment, not the landlord associated with the apartment
- Cards say "1 Review" instead of "1 Reviews"
- Clicking on a card or a search result routes to the apartment page, i.e. `apartment/:aptId` instead of the landlord page, i.e. `landlord/:landlordId`
- Some minor code clean up: destructuring in `Autocomplete.tsx`

Note: Many cards say "0 Reviews" because these apartments don't have any reviews, and the apartments that have reviews are not shown because they don't happen to be one of the first 3 (home page) or first 12 (reviews page) buildings in the `buildings` collection. We can address this in another PR, i.e. show most reviewed apartments. 

https://user-images.githubusercontent.com/57203639/144352479-d2bf78dd-f8c0-40a2-8ada-861383d17b04.mov

### Test Plan 
- Click on an apartment card from the `/reviews` page and home`/` page. Test that it goes to `apartment/:aptId`. 
- Click on a search result. Test that it goes to `apartment/:aptId`. 
- Check that addresses do not appear on a search result if the name of the building is the same as the address, e.g. 312 College Ave.


kjasani111 (approved):
These changes look good to me! I liked your code style that uses map to retrieve apartment data to prevent redundant code. I also liked how your code is very readable, as the names are all short and intuitive, and your changes that clean up the code in this PR actually made it more short and readable, yet still fully functional. I also tested that the user is redirected correctly to the apartment page instead of the landlord page from both the home and the reviews page, and the search correctly doesn't show the address for buildings with the same name as the address.

Issue #103, opened 2021-11-08
Populate apartment header and side panel info 


samliu10:
On `Apartment.tsx`, populate the following: 
- Details in the apartment header, e.g. name, average rating
- Details on the side info panel, e.g. landlord name, landlord contact, etc

PR #132, opened 2021-12-07
Implement Apartment Info Section and Header

menaattia:
### Summary 

This pull request implements the Header and Info Section on the Apartment page (review page for a given apartment)

- [x] closes issue #103 
- [x] implemented the Header for the Apartments, which is mostly similar to the Header component implemented for reviews by landlord, with changes made where needed
- [x] implemented the Info Section on the right side of the Apartment page, which is also the same as the landlord page but with info associated with apartment rather than landlord: landlord, contact, and address
- [x] made URLs of websites clickable in the Info section

### Test Plan 
Go to route `apartment/:id` using any valid apartment id from the database. ids: e.g. id 1, 2, 3, 7. The name of the apartment should be in the header instead of the name of the landlord/renting company. The average rating should be the average rating of the apartment, not the landlord. The info section should have the landlord/renting company, contact, address if they are available. The contact should be a clickable URL that redirects to the website of the renting company associated with that apartment. 

### Notes 
- A lot of code is copy pasted from the Landlord page. I could have spent time factoring out the code to make it reusable for both ApartmentPage and LandlordPage. However, for now it may not be necessary since we haven't decided if we want to keep the LandlordPage or have the ApartmentPage replace it.
- In the info section, the whole line "contact: [url]" is underlined because I'm using the same implementation as before which uses Material-UI's ListItemText and wrapped a <a> tag around it. If I wanted "contact" to not be underlined, I would have to change the implementation and styling.


frontend/src/components/Apartment/Info.tsx
```diff
@@ -0,0 +1,31 @@
...
+);
+
+export default function Info({ landlord, contact, address }: Props): ReactElement {
+  return (
+    <Box mt={1}>
+      <Typography variant="h5">Info</Typography>
+      <List dense>
+        {landlord && <InfoItem text={`Landlord/Renting Company: ${landlord}`} />}
+        {contact && (
+          <a href={contact}>
```
samliu10, line 23: Can we have this open an external tab? That way users don't lose their spot on our page

samliu10, line 23: Also, in the future we may want to change the color of the link. We don't have blue anywhere else on the site. It's a question for a designer though, so let's leave it as is for now.

samliu10 (approved):

samliu10 (approved):
lgtm

Issue #121, opened 2021-11-23
Show 404 page for apartments that don't exist


samliu10:
If a user navigates to `apartments/:aptId` where `aptId` is not a building that exists in the DB `buildings` collection, the linear progress bar spins forever. We should show the 404 page if a user tries to see reviews for an apartment that doesn't exist.

Issue #107, opened 2021-11-09
Website tab details 


samliu10:
Bugs identified from mvp launch bug bash
- Website title says “CU Apartments”: change to "CUAPTS" 
- Favicon (website tab icon) is still the old one

Issue #106, opened 2021-11-09
Footer fixes 


samliu10:
Bugs identified from mvp launch bug bash
- “Got a question that isn’t answered?” should be changed to "Want to get in touch?" 
- Privacy policy doesn't look clickable.

PR #137, opened 2021-12-19
Tab and footer

kjasani111:
### Summary 
This pull request implements showing a 404 page to the user if they navigate to a url for a landlord that doesn't exist, or a url for an apartment that doesn't exist. It also implements some small bug fixes.
- [x] Closes issue #106 
- [x] Closes issue #107 
- [x] Closes issue #121
- [x] Changed the styling for the privacy policy to make it look clickable by making it orange
- [x] Changed the wording in the footer from “Got a question that isn’t answered?” to "Want to get in touch?"
- [x] Changed the website favicon to the latest one
- [x] Changed the website tab title to "CUAPTS"
- [x] Implemented showing a 404 page instead of spinning forever if the user navigates to an invalid landlord or apartment id 
### Test Plan 
Run `yarn start` in the root folder. Check that the tab title, favicon, footer wording, and privacy policy is correct on the website.
- Try navigating to a landlord by clicking on their card from the home page. Confirm that the landlord page loads normally.
- Change the url to `landlord/landlord_id`, where `landlord_id` is a non-existent landlord. Confirm that the 404 page loads instead of the progress bar spinning forever.


frontend/src/pages/LandlordPage.tsx
```diff
@@ -48,18 +54,21 @@ const LandlordPage = (): ReactElement => {
   useEffect(() => {
     get<ReviewWithId[]>(`/review/landlordId/${landlordId}`, {
       callback: setReviewData,
+      errorHandler: handlePageNotFound,
```
samliu10, line 57: Same thing here- the error handler shouldn't be used for the review or buildings useEffect.

frontend/src/pages/ApartmentPage.tsx
```diff
@@ -60,18 +66,21 @@ const ApartmentPage = (): ReactElement => {
   useEffect(() => {
     get<ReviewWithId[]>(`/review/aptId/${aptId}`, {
       callback: setReviewData,
+      errorHandler: handlePageNotFound,
```
samliu10, line 69: The error handler should not be included in the `get<ReviewWithId[]>(/review/aptId/${aptId}` or  `get<Landlord>(/landlord/${apt?.landlordId}` useEffect since the 404 should only show if the `aptId` is not correct. If you test this with any route, e.g. `/apartment/10`, it goes to 404 when it shouldn't because 10 is a valid `aptId` in the `buildings` collection.

frontend/src/pages/LandlordPage.tsx
```diff
@@ -48,18 +54,21 @@ const LandlordPage = (): ReactElement => {
...
     get<Landlord>(`/landlord/${landlordId}`, {
       callback: setLandlordData,
+      errorHandler: handlePageNotFound,
     });
   }, [landlordId]);
 
   useEffect(() => {
     get<Apartment[]>(`/buildings/${landlordId}`, {
       callback: setBuildings,
+      errorHandler: handlePageNotFound,
```
samliu10, line 71: and here

frontend/src/pages/LandlordPage.tsx
samliu10, line 57: Same thing here- the error handler shouldn't be used for the review or buildings useEffect.

frontend/src/pages/ApartmentPage.tsx
samliu10, line 69: The error handler should not be included in the `get<ReviewWithId[]>(/review/aptId/${aptId}` or `get<Landlord>(/landlord/${apt?.landlordId}` `useEffect` since the 404 should only show if the `aptId` is not correct. If you test this with any route, e.g. `/apartment/10`, it goes to 404 when it shouldn't because 10 is a valid `aptId` in the buildings collection.

samliu10 (changes requested):

samliu10:
Good work! Just a few things to fix. Also for future reference, this probs should've been broken up into two PRs- one for the 404 and one for the tab/footer.

frontend/src/pages/ApartmentPage.tsx
kjasani111: I modified the code to fix this bug. Take a look when you get a chance!

kjasani111:
Hey @samliu10 I implemented the requested changes so it’s ready to be reviewed again. I’m not sure why but GitHub isn’t letting me re-request a review from you, which is why I’m messaging here instead.

frontend/src/pages/ApartmentPage.tsx
```diff
@@ -41,15 +42,23 @@ const ApartmentPage = (): ReactElement => {
   const [showSignInError, setShowSignInError] = useState(false);
   const [sortBy, setSortBy] = useState<Fields>('date');
   const toastTime = 3500;
-
+  const [notFound, setNotFound] = useState(false);
+  const [notFoundLandlord, setNotFoundLandlord] = useState(0);
+  const handlePageNotFound = () => {
```
samliu10, line 47: So actually, I realized the real reason why it's not working rn is because `handlePageNotFound` actually never gets called if `aptId` is not a valid id. This line needs to be added in `app.ts` so that we return an error if we encounter an invalid `aptId`. At the moment, if `aptId` is invalid, this route just returns a JSON object containing that id so the code is never reaching the `catch` clause. 

Basically, if the `aptId` is invalid, we should not even be looking for the landlord. We can actually get rid of everything related to `setNotFoundLandlord`. 
Screen Shot 2021-12-19 at 12 02 51 PM

frontend/src/pages/ApartmentPage.tsx
```diff
@@ -250,7 +256,9 @@ const ApartmentPage = (): ReactElement => {
     </Grid>
   );
 
-  return !loaded ? (
+  return notFoundLandlord === 2 || notFound ? (
```
samliu10, line 259: Then this will just become `return notFound ? (...)`

frontend/src/pages/ApartmentPage.tsx
```diff
@@ -41,15 +42,23 @@ const ApartmentPage = (): ReactElement => {
...
   const [sortBy, setSortBy] = useState<Fields>('date');
   const toastTime = 3500;
-
+  const [notFound, setNotFound] = useState(false);
+  const [notFoundLandlord, setNotFoundLandlord] = useState(0);
+  const handlePageNotFound = () => {
+    setNotFound(true);
+    setNotFoundLandlord(0);
+  };
+  const handleLandlordNotFound = () => {
```
samliu10, line 51: Can remove this

samliu10 (changes requested):

frontend/src/pages/ApartmentPage.tsx
kjasani111: Ok awesome! I thought that was the case so I came up with a patchwork solution, but I think this solution would work better. I'll change it.

kjasani111:
@samliu10 I implemented the requested changes so it's ready for review again!

samliu10 (approved):
LGTM!

PR #143, opened 2022-02-02
Other properties owned on apartment page

### LINKED ISSUE #140: Related properties on side panel
The side panel has a horizontal line below which used to be a section called "Properties owned" that listed the other properties owned by a specific landlord. Now that reviews are shown by apartment not landlord, that section is no longer there. We can repopulate it with a bulleted list of properties owned by the same landlord. 
Screen Shot 2022-01-28 at 3 15 23 PM

samliu10:
### Summary 

This PR closes #140 by populating informational side bar on the apartment page with the names of other properties owned by the same landlord. 
Screen Shot 2022-02-01 at 8 33 18 PM

### Test Plan 

- Navigate to `http://localhost:3000/apartment/211` or an apartment of your choice. 
- Check that the side bar shows the names of other properties 
- If the landlord only owns that one apartment, then it should just say "Other Properties Owned" with nothing underneath.

### DIFF
--- a/frontend/src/components/Apartment/AptInfo.tsx
+++ b/frontend/src/components/Apartment/AptInfo.tsx
@@ -1,19 +1,22 @@
 import React, { ReactElement } from 'react';
 import Info from './Info';
 import { Card, CardContent, Divider } from '@material-ui/core';
+import PropertyInfo from '../Review/PropertyInfo';
 
 type Props = {
   readonly landlord: string;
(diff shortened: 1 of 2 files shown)


thuypham03:
The sidebar of other properties owned by the same landlord makes users easier to explore if they want to find similar appartments and also encourages them to further explore our website. Great idea!

thuypham03 (approved):

PR #144, opened 2022-02-02
Display average rating number

### LINKED ISSUE #141: Display average review rating in page header
The header only shows the heart rating, not the number. 
Screen Shot 2022-01-28 at 3 19 01 PM

We can add the numerical value on the side, perhaps similarly to how other review based platforms do it. Ex: amazon (see below) 

Screen Shot 2022-01-28 at 3 19 58 PM

samliu10:
### Summary 

This PR closes #141 by: 
- Adds the average rating number in the header for an apartment
- Removes the empty hearts if an apartment does not have any reviews — this is consistent with other review-based platforms such as Amazon
- Changes the text to say "0 Reviews" instead of "0 Review"
Screen Shot 2022-02-01 at 9 38 02 PM
Screen Shot 2022-02-01 at 9 38 30 PM

### Test Plan 

- Navigate to `http://localhost:3000/apartment/200`: check that the empty hearts do not appear
- Navigate to `http://localhost:3000/apartment/210`: check that the average rating appears next to the hearts.

### DIFF
--- a/frontend/src/components/Apartment/Header.tsx
+++ b/frontend/src/components/Apartment/Header.tsx
@@ -8,6 +8,7 @@ import {
   withStyles,
   makeStyles,
   Avatar,
+  Typography,
 } from '@material-ui/core';
 import styles from './Header.module.scss';
 import { ApartmentWithId } from '../../../../common/types/db-types';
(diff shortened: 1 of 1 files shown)


thuypham03 (approved):
This change has uniformed our review session with other familiar review pages. Great job!

Issue #128, opened 2021-12-07
Optimize apartment search 


samliu10:
After writing a bunch of buildings to the `buildings` collection, the search bar is now really slow. It sometimes freezes for a very short amount of time because it's trying to filter through so many buildings. We need to optimize the search algorithm defined in the `/search` route in `backend/src/app.ts`. 

Edit: we need a new algorithm that takes into account the fact that now, the search bar is included in the nav bar

PR #149, opened 2022-03-06
Optimizing homepage search bar

thuypham03:
### Summary 

This PR optimizes the homepage search bar by:
- Retrieving all of the data (landlords and apartments) into global variables when first loading homepage rather than each time searching
- Show 5 most-matched results rather than every result

### DIFF
--- a/backend/src/app.ts
+++ b/backend/src/app.ts
@@ -139,8 +139,8 @@ app.get('/search', async (req, res) => {
       keys: ['name', 'address'],
     };
     const fuse = new Fuse(aptsLandlords, options);
-    const results = fuse.search(query);
-    const resultItems = results.map((result) => result.item).slice(0, 5);
+    const results = fuse.search(query).slice(0, 5);
+    const resultItems = results.map((result) => result.item);
 
     const resultsWithType: (LandlordWithLabel | ApartmentWithLabel)[] = resultItems.map((result) =>
       isLandlord(result)
--- a/backend/src/app.ts
+++ b/backend/src/app.ts
@@ -131,14 +131,8 @@ const isLandlord = (obj: LandlordWithId | ApartmentWithId): boolean => 'contact'
 app.get('/search', async (req, res) => {
   try {
     const query = req.query.q as string;
-    const landlordDocs = (await landlordCollection.get()).docs;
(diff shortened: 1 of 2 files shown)


samliu10 (commented):
Niice job! I tried it out, and the search is much faster and there is no lag. However, the code right now assumes that the search bar is only on the home page. I think Huey and Heewon were talking about having the search bar in the nav bar on all pages, so let's wait on this until we figure out if we are doing that.

thuypham03:
issue #128

backend/src/app.ts
```diff
@@ -168,6 +162,19 @@ app.get('/page-data/:page', async (req, res) => {
     (doc) => ({ id: doc.id, ...doc.data() } as ApartmentWithId)
   );
 
+  if (page === 'home') {
```
samliu10, line 165: Hmm so this approach works if the search bar is only on the home page. Now that we're adding it to the nav bar which appears on every page, we should try a different approach. For example, if someone directly navigates to an apartment page instead of first going to the home page, this part of the code would never be called. 

Just a thought: what if we check if the`landlords` and `apts`variables have a value and if not, we set them? Open to other ideas though!

backend/src/app.ts
```diff
@@ -131,21 +131,15 @@ const isLandlord = (obj: LandlordWithId | ApartmentWithId): boolean => 'contact'
...
+    const landlords = req.app.get('landlords');
+    const apts = req.app.get('apts');
     const aptsLandlords: (LandlordWithId | ApartmentWithId)[] = [...landlords, ...apts];
 
     const options = {
       keys: ['name', 'address'],
     };
     const fuse = new Fuse(aptsLandlords, options);
-    const results = fuse.search(query);
+    const results = fuse.search(query).slice(0, 5);
```
samliu10, line 142: This works for now. Later we can change this to include more results and just use pagination.

jiayuangu762943 (commented):
That's a great change and does speed up the search! I agree with Sam that since we are the search bar to the nav bar for every page, we should figure out new ways to handle all scenarios. Hope to discuss and learn more about pagination for the next work sesh!

samliu10:
Closed by #171

PR #163, opened 2022-03-19
blank search results page

jiayuangu762943:
Summary
This pull request is the first step towards implementing feature [SearchResultsPage]
As soon as a user press enter in the search bar, it will direct to a now blank search results page.

Test Plan
Not sure how to fully test in React, but a few manual checks confirm that it can direct to a blank page.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -1,6 +1,6 @@
 import React, { ReactElement } from 'react';
 import './App.scss';
-import { BrowserRouter as Router, Route, Routes } from 'react-router-dom';
+import { BrowserRouter as Router, Route, Switch } from 'react-router-dom';
 import HomePage from './pages/HomePage';
 import FAQPage from './pages/FAQPage';
 import ReviewPage from './pages/ReviewPage';
@@ -56,6 +56,11 @@ export type CardData = {
   company?: string;
 };
 
+export type LocationCardData = {
+  photo: string;
+  location: string;
+};
+
 const headersData = [home, review];
 
 hotjar.initialize(HJID, HJSV);
@@ -66,16 +71,16 @@ const App = (): ReactElement => {
       <Router>
         <NavBar headersData={headersData} />
         <div className="root">
-          <Routes>
-            <Route path="/" element={<HomePage />} />
-            <Route path="/faq" element={<FAQPage />} />
-            <Route path="/reviews" element={<ReviewPage />} />
-            <Route path="/policies" element={<Policies />} />
-            <Route path="/landlord/:landlordId" element={<LandlordPage />} />
-            <Route path="/apartment/:aptId" element={<ApartmentPage />} />
-            <Route path="/notfound" element={<NotFoundPage />} />
-            <Route path="/search" element={<SearchResultsPage />} />
-          </Routes>
+          <Switch>
+            <Route exact path="/" component={HomePage} />
+            <Route exact path="/faq" component={FAQPage} />
+            <Route exact path="/reviews" component={ReviewPage} />
+            <Route exact path="/policies" component={Policies} />
+            <Route path="/landlord/:landlordId" component={LandlordPage} />
+            <Route path="/apartment/:aptId" component={ApartmentPage} />
+            <Route path="/search" component={SearchResultsPage} />
+            <Route component={NotFoundPage} />
+          </Switch>
         </div>
         <Footer />
       </Router>
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -14,6 +14,7 @@ import { hotjar } from 'react-hotjar';
 import { HJID, HJSV } from './constants/hotjar';
 import Policies from './pages/Policies';
 import ApartmentPage from './pages/ApartmentPage';
+import SearchResultsPage from './pages/SearchResultsPage';
 
 const theme = createMuiTheme({
   palette: {
@@ -72,7 +73,8 @@ const App = (): ReactElement => {
             <Route exact path="/policies" component={Policies} />
             <Route path="/landlord/:landlordId" component={LandlordPage} />
             <Route path="/apartment/:aptId" component={ApartmentPage} />
-            <Route component={NotFoundPage} />
+            <Route exact path="/notfound" component={NotFoundPage} />
+            <Route exact path="/searchresults" component={SearchResultsPage} />
           </Switch>
         </div>
         <Footer />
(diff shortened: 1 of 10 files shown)


frontend/src/components/Home/Autocomplete.tsx
```diff
@@ -15,6 +15,7 @@ import { LandlordOrApartmentWithLabel } from '../../../../common/types/db-types'
 import SearchIcon from '@material-ui/icons/Search';
 import { makeStyles } from '@material-ui/core/styles';
 import { Link as RouterLink } from 'react-router-dom';
+import { useHistory } from 'react-router-dom';
```
samliu10, line 18: Turns out `useHistory` is deprecated in the newest version of `react-router-dom`. Let's do the following: 
- Upgrade to the latest version of `react-router-dom`: in the `frontend` directory, run `yarn add react-router-dom`
- Replace `useHistory` with `useNavigation`: see this [stack overflow post](https://stackoverflow.com/a/70379483) for how to use it
- Upgrading to this version will require other changes that I've noted in other comments

frontend/src/components/Home/Autocomplete.tsx
```diff
@@ -58,12 +59,17 @@ export default function Autocomplete() {
...
   const [selected, setSelected] = useState<LandlordOrApartmentWithLabel | null>(null);
   const [width, setWidth] = useState(inputRef.current.offsetWidth);
+  const history = useHistory();
 
   function handleListKeyDown(event: React.KeyboardEvent) {
     event.preventDefault();
     if (event.key === 'Tab') {
       setOpen(false);
     }
+    if (event.key === 'Enter') {
```
samliu10, line 69: This function `handleListKeyDown` is only for the `Menu` component so let's not make any changes to it

frontend/src/components/Home/Autocomplete.tsx
```diff
@@ -179,7 +185,16 @@ export default function Autocomplete() {
         placeholder="Search by landlord or building address"
         className={text}
         variant="outlined"
-        onKeyDown={(event) => (event.key === 'ArrowDown' ? setFocus(true) : setFocus(false))}
+        onKeyDown={(event) => {
```
samliu10, line 188: What do you think about moving this code to a separate function, similar to `handleListKeyDown`? I think that might be cleaner. Also do we need the `else` statement?

frontend/src/components/Home/Autocomplete.tsx
```diff
@@ -179,7 +185,16 @@ export default function Autocomplete() {
...
+          if (event.key === 'ArrowDown') {
+            setFocus(true);
+          } else if (event.key === 'Enter') {
+            setFocus(true);
+            history.push('/searchresults');
```
samliu10, line 193: So usually what happens when you search is that the URL will show the query parameters (the words after the `?`)- see the google image below. The query param is `q` and the value is `cats`. This way, a person can directly navigate to a page of search results, and this is more standard in industry. 
Screen Shot 2022-03-20 at 4 13 29 PM

How about we change `history.push('/searchresults');` to ```navigate(`/search?q=${query}`);```? Then the URL will show the query that the user types into the search bar

frontend/src/pages/SearchResultsPage.tsx
```diff
@@ -0,0 +1,18 @@
+import { Box, Container, Typography } from '@material-ui/core';
+import React, { ReactElement } from 'react';
+
+// import NotFoundIcon from '../assets/not-found.svg';
```
samliu10, line 4: Note for next time: always make sure to remove any commented code!

frontend/src/pages/SearchResultsPage.tsx
```diff
@@ -0,0 +1,18 @@
...
+// import styles from './NotFoundPage.module.css';
+
+const SearchResultsPage = (): ReactElement => {
+  return (
+    <Container maxWidth="md">
+      <Box pb={3} textAlign="center">
+        {/* <img src={NotFoundIcon} alt="CU Apts Logo" className={styles.NotFoundImage} /> */}
+        <Typography variant="h5">EMPTY NOW!</Typography>
```
samliu10, line 12: We will want to be able to access the query (what the user typed into the search bar) so that we can fetch the results for that query from the backend later. 

We can use `useSearchParams` from `react-router-dom` to access the query. See [this react documentation](https://reactrouter.com/docs/en/v6/api#usesearchparams) for how to use it.

samliu10: For now, we can make sure we are getting the correct query by displaying the result on the page
```
<Typography variant="h5">`EMPTY NOW! showing results for {getParams.get('q')}`</Typography>
```

frontend/src/App.tsx
```diff
@@ -72,7 +73,8 @@ const App = (): ReactElement => {
             <Route exact path="/policies" component={Policies} />
             <Route path="/landlord/:landlordId" component={LandlordPage} />
             <Route path="/apartment/:aptId" component={ApartmentPage} />
-            <Route component={NotFoundPage} />
+            <Route exact path="/notfound" component={NotFoundPage} />
```
samliu10, line 76: `<Switch />` in the newest version of `react-router-dom` is now deprecated, so you'll want to replace it with `<Routes />` which is documented on [this stack overflow post](https://stackoverflow.com/a/69849271). There are also some other syntax changes you'll have to make that the post describes ^^

frontend/src/App.tsx
```diff
@@ -14,6 +14,7 @@ import { hotjar } from 'react-hotjar';
 import { HJID, HJSV } from './constants/hotjar';
```
samliu10, line 14: Nice job! Some comments just to make the code more clean/updated :)

samliu10 (changes requested):

samliu10:
With those changes, we can get something that functions like this:

https://user-images.githubusercontent.com/57203639/159184754-7785e33d-7542-4815-9271-eb5af478ba60.mov

frontend/src/App.tsx
```diff
@@ -72,7 +73,8 @@ const App = (): ReactElement => {
             <Route exact path="/policies" component={Policies} />
             <Route path="/landlord/:landlordId" component={LandlordPage} />
             <Route path="/apartment/:aptId" component={ApartmentPage} />
-            <Route component={NotFoundPage} />
+            <Route exact path="/notfound" component={NotFoundPage} />
+            <Route exact path="/searchresults" component={SearchResultsPage} />
```
samliu10, line 77: Also, how about changing this to 
```  <Route path="/search" element={<SearchResultsPage />} />```? That way our search route is more similar to what people do in industry!

Issue #161, opened 2022-03-18
Standardize global styling: font, colors 


samliu10:
Task: standardize global styling (font, colors) across the project 
- Set global font
- Create a `colors.ts` file that defines and exports all colors used in the project 
- Replace hex color values with the colors in `colors.ts` so that they can be easily changed

PR #166, opened 2022-03-24
Global styling 

samliu10:
### Summary 

This PR is the first step in working towards #161 . These changes help standardize the styling across the project, making it easier to make changes and keep the styling consistent. 
- `colors.ts` defines the official exported variables for product colors. If at any time a color needs to be changed, the change only needs to be made in this file for it to be reflected across the website. 
- `createMUITheme` is now deprecated, so this changes all instances to `createTheme`
- The theme in `App.tsx` defines global styling for the entire website, in particular font sizes used in `<Typography />` components
- Defines the font used in the project as  `work-sans` 

From now on, all colors should be defined using `colors.ts` and all font sizes using a `<Typography variant="" />` unless custom styling is absolutely necessary. 

** Note: some of the styling may look odd at the moment now that we have this global styling that overrides styling in individual components. Those components may need to be changed later as necessary**  

### Test Plan 
- In the root directory, run `yarn start`
- Check that the project looks ok


cyrus14 (dismissed):
Hey Sam! I think this PR will be very useful to make future changes to the website a lot easier and cleaner. I think the site looks exactly like it's supposed to with your changes that you added. I did not see anything out of the ordinary. Looking through the code edited I see that most of the files have been transferred over to this new way of defining the colors. Do you think the color and style of ApartmentCards should also be adapted to this new standard? This is a small change so I will go ahead and approve, but just wanted to put that out there. Again, great job!

jiayuangu762943 (approved):
Great that we have standardized styling throughout making development cleaner and easier! That's a lot of work to refactor codes everywhere! Thank you so much Sam! So from now on when we are implementing cards we should use the global color right?

thuypham03 (approved):
Thanks Sam for creating the standardized global styling which makes our work much easier. You have done a great amount of work. Great job!

PR #171, opened 2022-04-15
Optimize search bar

jiayuangu762943:
Summary 

This pull request optimizes search by loading all the apartments and landlords data into global variables once when the page is loaded. The search bar will only query the stored global variables, avoiding sending requests to database every time one types a letter.

Test Plan 

Run `yarn start` and the responsiveness of search bar options pop-up is increased.


backend/src/app.ts
```diff
@@ -127,10 +127,8 @@ app.post('/new-landlord', async (req, res) => {
 });
 
 const isLandlord = (obj: LandlordWithId | ApartmentWithId): boolean => 'contact' in obj;
-
-app.get('/search', async (req, res) => {
+app.get('/getAllData', async (req, res) => {
```
samliu10, line 130: Sorry I didn't think about this earlier, but what do you think about changing this to a `POST` request? Technically we are just asking the backend to do something, and the frontend is not "getting" any data
```
app.post('/set-data', async (req, res) => {...})
```

frontend/src/App.tsx
```diff
@@ -66,6 +68,12 @@ const headersData = [home, review];
 hotjar.initialize(HJID, HJSV);
 
 const App = (): ReactElement => {
+  useEffect(() => {
+    get<any>(`/getAllData`, {
+      callback: () => {},
+    });
+  }, []);
+
```
samliu10, line 76: Great job!! Searching is so much faster this way, and it will definitely reduce the load on the database

samliu10 (changes requested):

backend/src/app.ts
samliu10: Also it is best practice in API design for routes to use `-` instead of `camelCase`, e.g. `/set-data` instead of `/setData`! Just a small thing

backend/src/app.ts
```diff
@@ -139,13 +137,28 @@ app.get('/search', async (req, res) => {
     const apts: ApartmentWithId[] = aptDocs.map(
       (apt) => ({ id: apt.id, ...apt.data() } as ApartmentWithId)
     );
+    app.set('landlords', landlords);
+    app.set('apts', apts);
+
+    const v: any = {};
```
samliu10, line 143: The backend doesn't need to send any data back to the frontend in this case, so you actually don't even need `v` here. That also means you can change the response code to `204` which means `No Content` in API design: see this example [here](https://www.rockyourcode.com/express.js-send-responses/)

samliu10 (approved):

backend/src/app.ts
samliu10: Oops accidentally pressed approve instead of request changes

samliu10 (changes requested):

samliu10 (approved):

PR #175, opened 2022-04-30
Landlord card

jiayuangu762943:
Summary 

This pull request is the first step towards populating the search results page. It shows landlord cards in a row.

https://user-images.githubusercontent.com/100146921/166978059-5990a1ae-2fbd-4c33-91a5-91e1c4cc343d.mov

Test Plan 

- At most three landlord cards is shown for a search query 
- if no landlord is returned by the search query, no landlord card is shown


frontend/src/components/LandlordCard/LandlordCard.tsx
```diff
@@ -0,0 +1,73 @@
...
+  card: {
+    height: '100%',
+    borderRadius: '10px',
+    backgroundColor: colors.landlordCardRed,
+    paddingTop: '10px',
+  },
+});
+
+const LandlordCard = ({ landlordData }: Props): ReactElement => {
+  const classes = useStyles();
```
samliu10, line 43: You can actually destructure this so that you don't have to keep using `classes.`! 
```
const {card, marginTxt, ... } = useStyles();
```

frontend/src/components/LandlordCard/LandlordCard.tsx
```diff
@@ -0,0 +1,73 @@
...
+  //useMediaQuery here is for detecting whether the screen size wider than 960px
+  // if so, matches is true; otherwise, it's false
+  const matches = useMediaQuery('(min-width:960px)');
```
samliu10, line 46: Maybe we should ask @shelost about this — should the image disappear on mobile?

samliu10 (changes requested):
Nice job! On figma, I see this as the mobile version: 
Screen Shot 2022-04-30 at 11 09 20 AM


I'm wondering if we should have each card take up the full width instead of putting all three next to each other. Otherwise, the font size will need to shrink since right now the landlord names overflow and get cut off. There may need to be some frontend fixes based on what Heewon says about the designs and how the components should resize when the screen size changes, e.g. define more breakpoints, deal with the size of the image, etc. Let's get some clarification on that!

samliu10 (approved):
LGTM

huyl3:
Congrats 🎉

PR #177, opened 2022-05-01
Changing other properties in Apartment Page

thuypham03:
[ONGOING] Successfully import ApartmentCards component into the other property. Further action:
- Modifying backend so the route '/buildings/:landlordId' return CardData[] instead of Apartment[]
- Implement breakpoints

### DIFF
--- a/frontend/src/components/Review/PropertyInfo.tsx
+++ b/frontend/src/components/Review/PropertyInfo.tsx
@@ -1,9 +1,11 @@
 import React, { ReactElement } from 'react';
 import Grid from '@material-ui/core/Grid';
 import { Box, Typography, List, ListItem, ListItemText } from '@material-ui/core';
+import { CardData } from '../../App';
+import ApartmentCards from '../ApartmentCard/ApartmentCards';
 
 type Props = {
-  readonly info: readonly string[];
+  readonly info: CardData[];
   readonly title: string;
 };
 
@@ -14,13 +16,7 @@ export default function PropertyInfo({ info, title }: Props): ReactElement {
       <List dense component="ul">
         <Grid container spacing={0} direction="row">
           {info.length === 0 && <Typography variant="body1">No information available.</Typography>}
-          {info.map((feature, index) => (
-            <Grid item xs={6} sm={12} md={6} key={index}>
-              <ListItem disableGutters>
-                <ListItemText primary={feature} />
-              </ListItem>
-            </Grid>
-          ))}
+          <ApartmentCards data={info} />
         </Grid>
       </List>
     </Box>
--- a/frontend/src/pages/ApartmentPage.tsx
+++ b/frontend/src/pages/ApartmentPage.tsx
@@ -19,6 +19,7 @@ import DropDown from '../components/utils/DropDown';
 import { useParams } from 'react-router-dom';
 import NotFoundPage from './NotFoundPage';
 import HeartRating from '../components/utils/HeartRating';
+import { CardData } from '../App';
 
 export type RatingInfo = {
   feature: string;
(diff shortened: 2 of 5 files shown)


frontend/src/components/Review/PropertyInfo.tsx
```diff
@@ -1,28 +1,112 @@
...
+  }, [id]);
+
+  const getAverageRating = (reviewData: ReviewWithId[]) =>
+    reviewData.reduce(
+      (currSum, { overallRating }) => (overallRating > 0 ? currSum + overallRating : currSum),
+      0
+    ) / reviewData.length;
+
+  return (
+    <Card className={classes.card}>
```
samliu10, line 54: Check the styling here on `PropertyCard`: why is there a huge gap on the page?
Screen Shot 2022-05-03 at 4 19 32 PM

frontend/src/components/Review/PropertyInfo.tsx
```diff
@@ -1,28 +1,112 @@
...
+            <Typography className={classes.aptNameTxt}>{name}</Typography>
+          </Grid>
+          <Grid container item justify="space-between">
+            <Grid>
+              <Typography variant="subtitle1">{address}</Typography>
+            </Grid>
+          </Grid>
+          <Grid container direction="row" alignItems="center">
+            <HeartRating value={getAverageRating(reviewData)} precision={0.5} readOnly />
+            <Typography className={classes.reviewNum}>{numReviews}</Typography>
```
samliu10, line 67: Make sure to add "Reviews" so that it looks like "2 Reviews" instead of "2"

samliu10 (changes requested):

backend/src/app.ts
```diff
@@ -114,6 +114,36 @@ app.get('/buildings/:landlordId', async (req, res) => {
...
 });
 
+app.get('/buildings/all/:landlordId', async (req, res) => {
+  const { landlordId } = req.params;
+  const buildingDocs = (await buildingsCollection.where('landlordId', '==', landlordId).get()).docs;
+  const buildings: ApartmentWithId[] = buildingDocs.map(
+    (doc) => ({ id: doc.id, ...doc.data() } as ApartmentWithId)
+  );
+
+  const pageData = await Promise.all(
```
samliu10, line 124: This entire code block for `pageData` is repeated in an existing route `app.get('/page-data/:page', async (req, res) => {...})`. Can you put it in a separate helper function and just call the function in both routes?

frontend/src/pages/ApartmentPage.tsx
```diff
@@ -115,6 +120,12 @@ const ApartmentPage = (): ReactElement => {
     return subscribeLikes(setLikedReviews);
   }, []);
 
+  useEffect(() => {
+    get<CardData[]>(`/buildings/all/${apt?.landlordId}`, {
+      callback: setOtherproperties,
```
samliu10, line 125: `otherProperties` is data that the page should have before the loading bar goes away. Can you change this `useEffect` on line 113 so that `setLoaded(true)` happens if `otherProperties` exists?
Screen Shot 2022-05-04 at 5 53 47 PM

samliu10 (changes requested):

samliu10 (approved):
LGTM! One thing to note for the future in case there are too many database reads, and you guys need to further optimize the backend/data retrieval — you can ask @shelost if he's fine with not including the number of reviews/the rating of the apartments in "other properties", i.e. only displaying the name and address of the apartment. Chances are, most people will not click on every card, and some people may not click on any cards! In that case, it is a waste of database reads to retrieve the review data for apartment cards that have not been clicked on.

huyl3:
🔥🎉

Issue #184, opened 2022-09-28
Search results page apartment cards


jiayuangu762943:
We want to populate the search results page with apartment cards that are returned by the search queries.
![Screen Shot 2022-09-27 at 9 50 56 PM](https://user-images.githubusercontent.com/100146921/192673858-be8489b5-d519-4cd6-924d-f3f04ad27322.png)

- [ ] create a branch `search-results-apartments`
- [ ] Remove the landlord in search results page `src/pages/SearchResultsPage.tsx`
- [ ] change the useEffect and useState `searchResults` hooks in search results page with a new return type 
- [ ] change the layout so that apartment cards are shown each in a row like homepage
- [ ] modify the backend search API handling in `backend/src/app.ts` in `app.get('/search')` so that the data sent back is the same as `app.get('/page-data/:page')` since this is the backend api used to render apartment cards on the home page. Refer to `pageData` async function to see how it's done.

PR #190, opened 2022-10-08
Removing Landlords and Adding Apt Cards on the Search Page (1/2)

juliejeong:
This PR is 1 of 2 parts addressing Issue #184. This PR removes the landlord cards on the search results and adds a placeholder text which will be replaced with apartment cards in the next PR #191 (2/2).

BEFORE:
image
AFTER:
image

### DIFF
--- a/frontend/src/pages/SearchResultsPage.tsx
+++ b/frontend/src/pages/SearchResultsPage.tsx
@@ -30,26 +30,15 @@ const SearchResultsPage = (): ReactElement => {
     });
   }, [query]);
 
-  // if properties is in searchItem, then it has the LandlordWithLabel type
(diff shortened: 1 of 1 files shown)


thuypham03:
Good job on your first CU-Apts PR! Thanks for joining our team. Beside removing the code for the card, please also remove unnecessary import/ variable (LandlordWithLabel, LandlordCard, searchResults) so that it can pass the deployment check above!

jiayuangu762943:
As Sophia pointed out, removing the unused imports and variables so that u can pass the CI test and increase the rendering speed. It's a good practice to just keep imports and variables you need in the code. Your task is a challenging one and that's a great start!!! : )

PR #195, opened 2022-10-14
Location Card Corresponding Title Image

cyrus14:
### Summary 

This Pull Request adds defining features such as an image and title for each location page so the user knows which location they are on.

This pull request is the first step towards implementing the full location page 

- implemented the Image and Title

Tested this new feature by launching it in the browser and changed screen sizes to make sure this new feature would work on all different screen sizes.

Screen Shot 2022-10-15 at 3 47 44 PM
Screen Shot 2022-10-15 at 4 02 12 PM


frontend/src/pages/HomePage.tsx
```diff
@@ -41,19 +45,19 @@ const HomePage = (): ReactElement => {
 
   const locationData: LocationCardData[] = [
```
jiayuangu762943, line 46: There was this `locationData` list defined already so that's reasonable to put the location-image mapping here. But  it's more straightforward logically to import the image and render them in `LocationCards`.  And from a performance point of view, importing unnecessary images high up in the ancestor and passing them as props both incur unnecessary latency. So consider switching this to `LocationCards`!

jiayuangu762943 (commented):
Hey Cyrus! Good job overall! Just checkout some of my comments in the code! Besides that, for the images of each location card, i think the corresponding images can be found in figma and you can export the images there. Let's try to export them in svg since it's more stable and put them in the assets folder!

jiayuangu762943 (approved):
LGTM!

PR #199, opened 2022-10-22
LocationPage with 2 Apartment Cards

cyrus14:
### Summary 
Added two apartment cards to the corresponding location page.

This pull request is the first step towards finishing the location page.

- [x] implemented an app.get request that filters by location and called inside of LocationPage

### Test Plan 
Tested new apartment cards in browser.

image


jiayuangu762943 (approved):
Good job!

frontend/src/pages/LocationPage.tsx
```diff
@@ -40,7 +37,9 @@ const useStyles = makeStyles({
   },
 });
 
-const LocationPage = ({ data }: Props): ReactElement => {
+const LocationPage = (): ReactElement => {
+  const [homeData, setHomeData] = useState<CardData[]>([]);
```
thuypham03, line 41: This variable should not be named `homeData` as it's the one for data in homepage. It would be better if you name it `aptCards` or just `data`

PR #200, opened 2022-10-24
pushing dans PR: Fix search result url link to apartment vs. landlord

cyrus14:
### Summary 
Previously, making a search query and clicking on a result brings the user to the page for an apartment, even if a landlord was clicked. This pull request fixes the issue by aligning the destination URL with the type of result: .org/apartment/21 vs. .org/landlord/21

### Test Plan 

image
image
image


jiayuangu762943 (approved):
Thx on behalf of Daniels!

Issue #185, opened 2022-10-05
inifinite scrolling on home page


jiayuangu762943:

jiayuangu762943:
Implement infinite scrolling on `ApartmentCards.tsx`

- create a branch called `infinite-scrolling-home-page`
- two great references https://codesandbox.io/s/yk7637p62z?file=/src/index.js:929-934 https://www.npmjs.com/package/react-infinite-scroll-component
- feel free to talk to me with any questions as this is completely new!

PR #201, opened 2022-10-28
Infinite Scrolling on Homepage

thuypham03:
This PR resolve issue #185 by adding infinite scrolling on homepage.

Steps: 
- Adding `loadingLength` in `frontend/src/constants/HomeConsts.ts`
- Modifying API `/page-data/:page/:size` and `/location/:loc/:size` in `backend/src/app.ts` by adding array size when loading data from database and returning `isEnded` (whether the collection query has been fully loaded)
- Create component `frontend/src/components/ApartmentCard/ScrollingCards.tsx` to handle the infinite scrolling, when reaching the bottom of the page, `dataSize` will be updated (`+ loadingLength`), causing the page to re-render to update the `data`
- Adding message 'Loading more apartments...' if it hasn't fully loaded, otherwise 'There's no more apartments!'

https://user-images.githubusercontent.com/62644357/198853212-2b91f330-b439-4c4e-8ca9-620bf7446366.mov

https://user-images.githubusercontent.com/62644357/202799047-7500fc62-e3e2-4924-ac60-26d54ead9d0d.mov


juliejeong (approved):
Nicely done! When tested locally, scrolling & loading more pages went smoothly and the codes look clean. One suggestion that could be added design-wise is maybe adding something like "load more" or some indication that there are more pages to be loaded. so that it is more clear & intuitive to the user since as of right now it's a bit subtle.

cyrus14 (commented):
Hi Sophia! Great job on this PR to add infinite scrolling functionality to the homepage. I really like how you added what Julie suggested which was to show a "load more" indicator so that the user is able to intuitively expect more apartments to load. I think it's also great how you've abstracted the size of how many apartments to retrieve so that it is more versatile for future features we would want to add. Last thing to add, I think you could remove the 'page' variable on line 206 in backend/src/app.ts because it is never used.

juliejeong (commented):
The new changes look great! Definitely more efficient and structured than the previous version. Scrolling looks flawless when tested locally! Nice job!

dan-ieljin:
This PR for adding infinite scrolling looks great! The load more functionality is definitely a smart choice since it does not overwhelm the user with content. No issues when testing on my local machine. The new API endpoint also is cleanly implemented with intuitive code and error handling.

cyrus14 (approved):
Hi Sophia! I'm reviewing this PR for the second time after seeing the changes you made to your original PR. Just to reiterate really good job creating the infinite scrolling. This is very beneficial for our platform as a whole since we are dealing with so many listings. The load more indicator is good for users because they will understand that by scrolling they can load more listings. Also great job on removing unused variables. Approved!

PR #191, opened 2022-10-08
[WIP] Removing Landlords and Adding Apt Cards on the Search Page (2/2)

juliejeong:
This PR is 2 of 2 parts addressing Issue #184 with PR #190.

[WORK IN PROGRESS]
The finished PR should look like this:
image

PR #202, opened 2022-10-28
Search Results Page with Apartment Cards (NEW implementation)

juliejeong:
### Summary 

This PR allows apartment results to be displayed when a user searches a keyword. The previous, now closed, PR #191, used the same `/search` route as before (which returns AptOrLandlord results) and filtered the returned results to contain only apartments, then required a complex type conversion. Now, this new PR creates a new route that retrieves only apartment data from the query and returns the results with all the required data in a form that can be directly used in frontend.

This PR does the following:
- remove the landlord cards from the SearchResultsPage
- create a new route `/search-results` that returns apartments that match the search query as a pageData object
- renders the returned results as ApartmentCards in SearchResultsPage

### Test Plan 

Tested locally by passing different queries into the search bar and examine the returned results.


juliejeong:
please refer to PR #207 for the modified version without the conflicts

Issue #215, opened 2023-02-08
Change infinite scrolling on homepage to table with fixed length


thuypham03:

cyrus14:
Finished this ticket!

PR #221, opened 2023-02-10
No Infinite Scroll On Homepage

cyrus14:
### Summary 

In response to https://github.com/cornell-dti/cu-apts/issues/215 , this commit makes it so there is no more infinite scroll on the homepage.

### Test Plan 

Tested by opening in browser to make sure there is no infinite scrolling on the homepage but there is infinite scrolling for all of the other pages.

image

### DIFF
--- a/frontend/src/components/ApartmentCard/ScrollingCards.tsx
+++ b/frontend/src/components/ApartmentCard/ScrollingCards.tsx
@@ -31,8 +31,6 @@ const ScrollingCards = ({ API }: Props): ReactElement => {
     });
   }, [dataSize, API]);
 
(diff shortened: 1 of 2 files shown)


AnkitLakkapragada:
Tested this fix and it looks really good! Great work, this is very beneficial for usability.

thuypham03 (approved):
Good job for reusing the existed component to finish task. It's the art of usability! One small suggestion is that you should prevent using magic numbers, but rather constant that will be used globally. It helps keeping track of our page better too! I have adapt your PR to constant this time, so no further action needed!

PR #224, opened 2023-02-16
Add 'status' property to Review type

dan-ieljin:
### Summary 

This pull request is the first step in preparing for an admin page and report abuse button.

- added 'status' property to Review type with possible values being one of PENDING, APPROVED, DECLINED
- added default status of PENDING when creating a new review in the '/new-review' endpoint

### Test Plan 
I created a test review and confirmed that the status property was set to PENDING in Firestore.


thuypham03 (approved):
Good job! I have tested out on Postman on adding new comment and it's in 'PENDING' state as we're all expected. Thanks for helping our team writing additional scripts to add 'state' to every existed reviews in firebase!

PR #232, opened 2023-03-03
Resolve miscellaneous warning when launching site

thuypham03:
When launching site and monitoring console (F12), many warnings occur and had been resolved by:
- Update `justify` to `justifyContent`
- Re-import Firebase SDK properly
- Remove type any in add data script


AnkitLakkapragada (approved):
Looks good! This change is important and it's good to know we were able to resolve those warnings. Are all the imports in `firebase.ts` used?

PR #233, opened 2023-03-03
Add User Log In Button to Nav Bar

cyrus14:
### Summary 

Implemented a Sign In/Sign Out Button for the Navbar.

If the user is not logged in, there is a login button (placed at right-end of the header) that allows user to sign in with Google Login (already implement with Firebase)

If the user is logged in, there is a logout button (placed at right-end of the header) that allows user to logout

When user tries to write a review they will be prompted with sign in. If they sign in through that route, the button will update accordingly.

### Test Plan 

Resized the browser to make sure it worked on screen sizes. Tested different ways of signing in to make sure that the states stayed consistent.

https://user-images.githubusercontent.com/60114462/222983958-c9279c80-b0b1-41d2-a3d0-7b51e4e1bc57.mov


frontend/src/pages/ApartmentPage.tsx
```diff
@@ -205,14 +213,15 @@ const ApartmentPage = (): ReactElement => {
...
+    let user = await getUser(true);
     if (!user) {
-      let user = await getUser(true);
       setUser(user);
       if (!user) {
         showSignInErrorToast();
         return;
       }
     }
+    setUser(user);
```
thuypham03, line 224: You always `setUser(user)` no matter the if statement above result in, please move this line up before if statement, then delete the other getUser

frontend/src/components/utils/NavBar/index.tsx
```diff
@@ -158,6 +186,27 @@ const NavBar = ({ headersData }: Props): ReactElement => {
     );
   };
 
+  const signInAction = async () => {
+    console.log(user);
```
thuypham03, line 190: Please remove this `console.log` debugging statement.

frontend/src/components/utils/NavBar/index.tsx
```diff
@@ -203,19 +252,18 @@ const NavBar = ({ headersData }: Props): ReactElement => {
...
-  const searchBar = location.pathname !== '/';
-
   const displayDesktop = (): ReactElement => {
     return (
-      <Grid container className={toolbar} alignItems="center">
+      <Grid container className={toolbar} alignItems="center" justifyContent="space-between">
         <Grid item md={3}>
           {homeLogo}
         </Grid>
-        <Grid item md={6} className={searchBar ? search : searchHidden}>
```
thuypham03, line 214: I believe this line is for adapting new design on homepage. The search bar in navigation page disappear when we're on homepage, otherwise it stays still. Please revert this back then add additional styling so the nav bar appear as expected!

thuypham03 (changes requested):
Please take a look at those comments!

frontend/src/components/utils/NavBar/index.tsx
```diff
@@ -141,6 +161,14 @@ const NavBar = ({ headersData }: Props): ReactElement => {
     setDrawerOpen(false);
   }, [location]);
 
+  useEffect(() => {
+    if (user) {
+      setButtonText('Sign Out');
+    } else {
```
thuypham03, line 167: For this part, please use conditional expression, like `setButtonText(!user ? 'Sign In' : 'Sign Out')`

frontend/src/components/utils/NavBar/index.tsx
```diff
@@ -248,18 +251,19 @@ const NavBar = ({ headersData, user, setUser }: Props): ReactElement => {
...
+            </Link>
           </Grid>
         </Grid>
       </Grid>
     </Grid>
   );
 
   const searchBar = location.pathname !== '/';
+  // const checkUser = () => {
+  //   console.log(isCurrentUser());
```
thuypham03, line 265: Please remove any code and debugging statement that you won't use.

thuypham03 (dismissed):
Look good to me at this moment! We will note on integrating token to check auth status and try change user from internal state to global context later!

thuypham03 (approved):
Look good to me at this moment! We will note on integrating token to check auth status and try change user from internal state to global context later!

PR #240, opened 2023-03-13
Implement front end UI for Admin page

AnkitLakkapragada:
### Summary 

This pull request is the first step towards implementing the admin page that approves and rejects reviewers for apartments.

The page can be accessed at ` /admincuapts1234`. The next steps include creating an API to fetch all pending reviews and giving this button functionality and this webpage security.  

### Test Plan 

image

### Breaking Changes  

 
None

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, useEffect } from 'react';
+import React, { ReactElement, useEffect, useState } from 'react';
 import './App.scss';
 import { BrowserRouter as Router, Route, Switch } from 'react-router-dom';
 import HomePage from './pages/HomePage';
@@ -89,6 +89,7 @@ const headersData = [home, faq];
 hotjar.initialize(HJID, HJSV);
 
 const App = (): ReactElement => {
+  const [user, setUser] = useState<firebase.User | null>(null);
   useEffect(() => {
     const setData = async () => {
       await axios.post('/set-data');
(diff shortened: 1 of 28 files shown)


cyrus14 (dismissed):
Loving this front end UI for Admin Page. I think this is a great foundation for the Admin Page. Once the backend is figured out I think some adjustments will need to be made to have the approve and reject be placed in a way that is aligned with the review that you are deciding for. For now though, this is great for the set up!

frontend/src/components/utils/NavBar/index.tsx
```diff
@@ -270,6 +305,9 @@ const NavBar = ({ headersData, user, setUser }: Props): ReactElement => {
         <Grid item md={6} className={searchBar ? search : searchHidden}>
           {auto()}
         </Grid>
+        <Grid item md={1} className={menu} container justifyContent="flex-end">
```
thuypham03, line 308: The logic will be better if you have another boolean isAdmin then you do {isAdmin && <Button>}, the button should only render the button, not to check whether it's admin or not

frontend/src/App.tsx
```diff
@@ -90,6 +90,19 @@ hotjar.initialize(HJID, HJSV);
 
 const App = (): ReactElement => {
   const [user, setUser] = useState<firebase.User | null>(null);
+  const admins: string[] = [
```
thuypham03, line 93: This should be moved to const file.

AnkitLakkapragada (commented):
Looks good, this change is really needed for the front-end UI and also the first step for the admin page that will allow us to approve reviews.

thuypham03 (approved):
This PR finish implementation for admin page UI. In the future, the button need to be functional to change review from PENDING to ACCPETED/ DECLINED!

PR #244, opened 2023-03-14
Change design in Review card (Reduce hearts)

thuypham03:
New design: Reduce the number of heart
Screenshot 2023-03-14 at 1 30 38 AM

Old design:
Screenshot 2023-03-14 at 1 32 44 AM


cyrus14 (approved):
Great job Sophia on reducing the number of hearts. It is much less overwhelming and is consistent with the color scheme!

PR #245, opened 2023-03-14
Change hard reload on logo to switch router

thuypham03:
Previously, when clicking logo, the page will hard reload. However, now it just change the component without render everything, which keep the login/logout status be more stable.


cyrus14 (approved):
Hey Sophia! This was a great work around for the issue we were having where hard reloading the webpage reset the user State! By not using the old link feature that completely reloaded the page, we now just call the React Route. Everything looks good to me!

PR #247, opened 2023-03-15
Activate report abuse button for reviews

dan-ieljin:
### Summary 

This pull request activates the Report Abuse button for individual reviews. After clicking, the review status in the database will change to PENDING, and the page will reload to show only approved reviews.

* Added button click handler

### Testing 
* Manual


thuypham03 (dismissed):
.

thuypham03 (approved):
Good job on handling the abuse button! In the future, we might migrate user from prop to context to conveniently use it rather than the way we did for adding authentication right now!

PR #248, opened 2023-03-19
Corrected the number of approved reviews on the Home page

jesshan700:
### Summary 
Corrected the number of approved reviews shown on the home page. Looked through backend code to get a better understanding of the app infrastructure. 

This pull request is the first step towards fixing the number of approved reviews on the home page. 

- [x] Fixed the number of approved reviews so that it shows the correct number of approved reviews on the home page.

### Test Plan 

Added a new review and manually changed the status from PENDING -> APPROVED to see if the number of approved requests were correct. 

### Breaking Changes  

### DIFF
--- a/backend/src/app.ts
+++ b/backend/src/app.ts
@@ -124,7 +124,10 @@ const pageData = async (buildings: ApartmentWithId[]) =>
         throw new Error('Invalid landlordId');
       }
 
-      const reviewList = await reviewCollection.where(`aptId`, '==', id).get();
+      const reviewList = await reviewCollection
+        .where(`aptId`, '==', id)
+        .where('status', '==', 'APPROVED')
+        .get();
       const landlordDoc = await landlordCollection.doc(landlordId).get();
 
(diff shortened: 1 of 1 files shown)


backend/src/app.ts
```diff
@@ -288,20 +291,4 @@ app.post('/add-like', authenticate, likeHandler(false));
 
 app.post('/remove-like', authenticate, likeHandler(true));
 
-app.put('/update-review-status/:reviewDocId/:newStatus', async (req, res) => {
```
thuypham03, line 291: Can you revert the deletion of this API, we will need this at other task!

thuypham03 (changes requested):

thuypham03 (approved):
Great job! I have tested it out and it works perfectly. Hope that this bug helps you understand better about CUApts codebase.

ggsawatyanon:
Already merged, but the backend logic looks great. This works as intended in testing.

Issue #241, opened 2023-03-13
Implement mobile display for Homepage


thuypham03:
Refer design: https://www.figma.com/file/rUX3wOgAaZ40GWf5HwMTkV/CUAPTS-SP'23?node-id=184%3A2800&t=PJdUIvfXqyffCVnj-0

PR #251, opened 2023-03-20
Implement mobile display for Homepage V2

cyrus14:
### Summary 

Made it so the apartment listing are removed on mobile, and made the location cards 2x2 for better accessibility to mobile users. Referred to figma for looks. Addresses ticket #241 

### Test Plan 

Resized the browser window for mobile display.

image


AnkitLakkapragada (commented):
These changes are very good and make the mobile display a lot better for the homepage. There may be a way to have the isMobile field duplicated so you can save on the creation of the variable. Overall changes look good!

PR #255, opened 2023-03-22
Fix screen freeze after signing in from report abuse button

dan-ieljin:
### Summary 

When a user is not signed in, clicking Report Abuse from an apartment/landlord page will prompt them to sign in. Previously, when returning to the page, it would freeze, and nothing would be clickable. 

* prompt for sign in
* force user to click Report Abuse again after signing if desired

### Test Plan 
After returning to page, user can successfully perform all available actions, such as leave a review
image


thuypham03 (approved):
LGTM! This change have catch one of the crucial bug after the user sign in to make the experience smoother!

PR #263, opened 2023-03-30
Add deployment build path for front end

thuypham03:


michelleli01 (approved):
hopefully this will serve ou our front end in heroku

PR #269, opened 2023-04-11
Mobile Location Page 

cyrus14:
### Summary 

This pull request is to make the location page more like the figma prototypes.

- removed "loading more apartments" text when on mobile.

### Test Plan 
Tested on different sized windows

image

image


thuypham03 (approved):
Thanks for reverting back normal apt card from scrolling card to reuse the show more button in the component. Clean code!

PR #287, opened 2023-05-05
Search bar query bug

jesshan700:
### Fixed search bar query bug

This pull request is the first step towards fixing the search bar query bux

- [x] fixed the search bar so that every time the user switched to a different page, the search bar is cleared.

### Test Plan 
**1) After searching "main" from the homepage**
Screen Shot 2023-05-05 at 1 37 26 PM

**2) Search results for bar - search bar is cleared**
Screen Shot 2023-05-05 at 2 44 25 PM

**3) Search new from the home screen - search bar is cleared**
Screen Shot 2023-05-05 at 2 45 11 PM

Screen Shot 2023-05-05 at 2 45 18 PM

**4) Switching to FAQ - search bar is cleared - before fix search query was saved in search bar**
Screen Shot 2023-05-05 at 2 45 55 PM

### Breaking Changes  

### DIFF
--- a/frontend/src/components/Home/Autocomplete.tsx
+++ b/frontend/src/components/Home/Autocomplete.tsx
@@ -82,6 +82,7 @@ export default function Autocomplete() {
     } else if (event.key === 'Enter') {
       setFocus(true);
       history.push(`/search?q=${query}`);
+      setQuery('');
       setOpen(false);
     }
   }
--- a/frontend/src/components/Home/Autocomplete.tsx
+++ b/frontend/src/components/Home/Autocomplete.tsx
@@ -83,8 +83,8 @@ export default function Autocomplete() {
     } else if (event.key === 'Enter') {
       setFocus(true);
       history.push(`/search?q=${query}`);
-      setQuery('');
       setOpen(false);
+      setQuery('');
     }
   }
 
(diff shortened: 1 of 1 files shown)


cyrus14 (approved):
Hey Jessica! Great job on this UI change this fixes the bug where the search bar was outlined. Looking at the code changes, the setQuery('') for both the events where the down arrow is pressed and enter is pressed is important to match the intuition of the user. All in all, great job on this PR and it is good to have it in the product!

AnkitLakkapragada (approved):
Great work Jessica! This is an important fix and bugs are always important to find and squash. Looking forward to seeing more!

thuypham03 (changes requested):
The fixing looks great. However, the bug of search bar not reset after switching between different pages still stay, please update this PR with all bugs related to this issue fixed before merge!

ggsawatyanon:
This PR looks great. However, apart from Sophia's review that the bug remains when switching pages, I also noticed that the search bar is cleared after the down arrow key is pressed, which may not be something the user wants. Also because the search query doesn't actually go through when the down key is pressed so it just clears it without searching anything. Still, good progress though!

thuypham03:
Fixed in #302

PR #294, opened 2023-09-06
Bug fix for mobile footer

AnkitLakkapragada:
### Summary 

- Bug fix for mobile footer to ensure it is flush with the bottom of all mobile screens

### Test Plan 

- testing using different devices

### Breaking Changes  


thuypham03:
This changed didn't fix the root problem of why footer is hanging. The problem is in css '.root', we set the body of the screen with height: `min-height: calc(100vh - 120px - 94px);` but the actual should just be  `min-height: calc(100vh - 94px);`. Check this PR: https://github.com/cornell-dti/cu-apts/pull/295

PR #297, opened 2023-09-24
Implement Delete Button to Admin Page

AnkitLakkapragada:
### Summary 

This pull request implements the delete button to the admin page for the declined section so that admin users can stop seeing reviews that they want to delete. 

PNG image
PNG image

### Test Plan 

Tested it locally.

### Breaking Changes  
None

### DIFF
--- a/backend/src/app.ts
+++ b/backend/src/app.ts
@@ -300,7 +300,7 @@ app.post('/api/remove-like', authenticate, likeHandler(true));
 
 app.put('/api/update-review-status/:reviewDocId/:newStatus', async (req, res) => {
   const { reviewDocId, newStatus } = req.params;
-  const statusList = ['PENDING', 'APPROVED', 'DECLINED'];
+  const statusList = ['PENDING', 'APPROVED', 'DECLINED', 'DELETED'];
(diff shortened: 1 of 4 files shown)


cyrus14:
Hey Ankit! Great job on this feature, I think you've made a valuable addition to our admin page with the delete button for the declined section. This enhancement will undoubtedly streamline the workflow for our admin users, making it easier for them to manage and declutter their review queues. Looking at the code I like how you don't have any redundancies and the functionality works great when I tested it. Again great job!

mluo24 (commented):
Code looks good, and this feature works! I especially like that you made it a status rather than actually deleting it, as soft deletes are generally good practice. I also like how you reused the components for showing the reviews and just added an extra argument. A good solution!

Not sure what's up with your yarn file though, since there shouldn't be any changes as far as I'm aware.

jesshan700:
This looks really good! The delete button is a really important feature when it comes to how users will view our product - it will definitely clean up a lot on the admin page.

thuypham03 (approved):
Good job on soft-delete practice!

PR #305, opened 2023-10-09
Differentiating landlord reviews

jesshan700:
### Added to landlord reviews 

This pull request is the first step towards showing the difference between landlord reviews and apartment reviews on the landlord page. Checks the aptId field for each review to see if it is empty or null indicating a landlord review.

- [x] Text that shows the user which reviews are landlord reviews

### Test Plan 

Testing Lambrou landlord vs apartment:
Screen Shot 2023-10-09 at 3 15 29 PM

Testing Travis Hyde landlord vs apartment:
Screen Shot 2023-10-09 at 3 18 27 PM

### Notes 

- Decided not to make the text a link because it could get confusing for the user if the link just directs them to the page that they are already on. 
- Could maybe implement some sort of icon to show the differentiation vs text!

### Breaking Changes  


frontend/src/components/Review/Review.tsx
```diff
@@ -162,6 +180,15 @@ const ReviewComponent = ({
                       <ExpandMoreIcon />
                     </IconButton>
                   </Grid>
+
+                  {/* Checking to see if aptId of the review is null or empty, indicating that the review is a 
+                  landlord apartment - if landlord apartment is true, will add landlord review text */}
+                  <Typography className={landlordIndicator}>
+                    {review.aptId === '' || review.aptId == null
```
mluo24, line 187: I think you can do `!review.aptId` instead.

Also, with what I was saying above, if you add `landlordName` as a prop, you can rewrite the code as such:

```
{landlordName && (
<Typography className={landlordIndicator}>
  'Landlord Review - ' + landlordName
</Typography>
)}
```

This also only renders the component if `landlordName` has an input.

frontend/src/components/Review/Review.tsx
```diff
@@ -99,14 +108,23 @@ const ReviewComponent = ({
...
     setExpanded(!expanded);
   };
 
+  //Retreieving landlord data
+  useEffect(() => {
+    get<Landlord>(`/api/landlord/${review.landlordId}`, {
+      callback: setLandlord,
+    });
+  }, [review]);
+
```
mluo24, line 127: I think this API call isn't actually necessary. The parent component already should have made this API call and acquired the data already (since it's the landlord page), and since you only need the name of the landlord, it would make sense to just pass the name of it as a prop to this component, especially since this component is used in other places where you don't need the landlord name.

I would suggest adding an optional prop to this component that passes in the landlord name, and in 188 replace the `landlord?.name` with this prop, and maybe change the condition so it relies on the landlord name being not null instead.

mluo24 (commented):
Nice work Jessica! You really made the indicator look nice, and it's clear which one is a landlord review. Maybe you can also add indication for what's an apartment review with the apartment name?

Your logic with your solution is great, but I left some comments for how to micro-optimize it.

jesshan700:
Updated the differentiation between landlord and apartments so that apartment links with the specific apartment shows on the card rather than showing the landlord review. Changed the styling so that it matches with the admin page.

Screen Shot 2023-10-16 at 4 06 49 PM


Redirects user to the specific apartment of that card. 

Screen Shot 2023-10-16 at 4 07 38 PM

jesshan700:
Screen Shot 2023-10-22 at 10 31 33 PM

Changed styling of properties indicator rather than red link as shown above, better indicator that the link represents the property. Underlined portion is a link to the respective apartment page.

ggsawatyanon (approved):
This looks good, Jessica! This is a really helpful feature for users to understand the UI of the landlord reviews page easier. I'm not sure the purpose of the addition of the install_nvm.s file though. Good job on this PR!

thuypham03 (approved):
Good job on adding a differentiation between landlord and apartment's review on landlord page. Now, it's easier for the user to follow and can easy redirect to the corresponding landlord page.

PR #309, opened 2023-10-22
Updating Apartment Page UI/Functionality V3

cyrus14:
### Summary 

This PR is to update the current User Interface of the apartment page. Instead of the linear scale for the feature ratings, there is now 5 subsections for each review scale.

### Test Plan 

Tested on resizing the browser window.
The Current Look:
image

Figma:
image


ggsawatyanon:
This PR looks great and implements a needed feature that provides users with more information about each apartment's reviews! Perhaps you could delete the handleShowMore function instead of commenting it out? Good job!

AnkitLakkapragada (commented):
Great work Cyrus! This code is very needed and makes the user experience a lot better. I also really liked the comments that you put throughout. Looking forward to seeing it in action!

thuypham03 (changes requested):
The new segment doesn't work on admin page due to some problem while rendering AdminReviewComponent --> LabeledLinearProgress

thuypham03 (approved):
Good job on adapting new review bar design! This PR makes the score much easier to estimate visually with separated segments. Thanks for also fixing bug in admin page which previous does not render the new review bar properly.

PR #314, opened 2023-11-06
implement top most reviewed building for home

mluo24:
### Summary 

Changes:

1. Modification of an existing function that previously fetched all apartments (now depends on the page).
2. On the home page, the top five apartments with the most reviews are displayed.

- The code takes a parameter `page` to determine the page to display.
- If the `page` is 'home', it retrieves all building documents and calculates the total number of approved reviews for each building.
- The building documents and their corresponding review counts are sorted in descending order based on the review counts.
- The top five buildings with the most reviews are selected for display on the home page.
- Everything else from before remains the same.

These changes improve the user experience by showcasing the top-rated apartments on the home page, making it easier for users to discover popular choices.

### Test Plan 

image

### Notes 

Maybe slow.


backend/src/app.ts
```diff
@@ -261,10 +261,42 @@ app.get('/api/search-results', async (req, res) => {
...
+              .where('aptId', '==', doc.id)
+              .where('status', '==', 'APPROVED')
+              .get()
+          ).docs.length,
+          doc,
+        ])
+      );
+      buildingReviewCounts.sort().reverse();
+      // get top five most reviewed
+      buildingDocs = buildingReviewCounts.splice(0, 5).map((elem) => elem[1]);
```
thuypham03, line 294: Can we try to reuse the `const pageData = async (buildings: ApartmentWithId[])` function which takes in `ApartmentWithId[]` and returns the corresponding data with the number of reviews and company? Later on, we still need to call this function (line 305) to retrieve the information. If we can call it before and use the information in that, we don't need to retrieve the `reviewCollection` twice.

thuypham03 (changes requested):

backend/src/app.ts
mluo24: ty, didn't realize this function existed!

thuypham03 (approved):
Good job! This helps improve the homepage a lot by showing our most popular apartments for suggestion.

PR #319, opened 2023-11-09
Add endpoint for retrieve liked reviews of current user

thuypham03:
Add endpoint GET `/api/review/like` that returns a list of `ReviewWithId[]` that the current user marked as helpful (like). This endpoint can only be called when the user has signed in. 

Testing: Postman with unauthenticated version `/api/review/like/:userId` as below:
- GET: http://localhost:8080/api/review/like/P6abmpm0yaclfZ2RWfokdbTuADD3
- Result (in the picture):  
Screenshot 2023-11-09 at 5 31 38 PM

### DIFF
--- a/backend/src/app.ts
+++ b/backend/src/app.ts
@@ -97,6 +97,7 @@ app.get('/api/review/like', authenticate, async (req, res) => {
       const matchingReviews: ReviewWithId[] = [];
       const querySnapshot = await reviewCollection
         .where(FieldPath.documentId(), 'in', reviewIds)
+        .where('status', '==', 'APPROVED')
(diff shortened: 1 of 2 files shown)


AnkitLakkapragada (dismissed):
Awesome work Sophia! These changes are very important for our backend. I also really like how detailed you were with the testing plan. One thing to consider is maybe maintaining the same case names, on line 23.

cyrus14 (approved):
Great job, Sophia! These modifications carry significant importance for our backend system. I particularly appreciate the thoroughness of your testing plan. Looking forward to seeing this in production!

PR #320, opened 2023-11-11
Delete review endpoint & Edit review endpoint

cyrus14:
### Summary 

- Added edit and delete review endpoints into app.ts which will be used for user profile page actions.
- Can now be called when a user edits or deletes their own review in their user profile

### Test Plan 

Tested endpoints with Postman

https://github.com/cornell-dti/cu-apts/assets/60114462/bb43e7d4-e53d-41b5-81bc-53f41a63fc9c


backend/src/app.ts
```diff
@@ -319,7 +319,48 @@ app.post('/api/add-like', authenticate, likeHandler(false));
 
 app.post('/api/remove-like', authenticate, likeHandler(true));
 
-app.put('/api/update-review-status/:reviewDocId/:newStatus', async (req, res) => {
+// Endpoint to update a review by its document ID
+app.put('/api/edit-review/:reviewDocId', async (req, res) => {
```
mluo24, line 323: Generally the convention is you get the id and then you do the action, so I think `:reviewDocId/edit` would be more apt.

Same with line 348

backend/src/app.ts
```diff
@@ -319,7 +319,48 @@ app.post('/api/add-like', authenticate, likeHandler(false));
...
+
+    // Check if the required fields (overallRating and reviewText) are missing or invalid
+    if (review.overallRating === 0 || review.reviewText === '') {
+      res.status(401).send('Error: missing fields');
+    }
+
+    // Update the review document in the database with the provided data
+    await reviewCollection
+      .doc(reviewDocId)
+      .update({ ...review, date: new Date(review.date), status: 'PENDING' });
```
mluo24, line 336: Are you leaving the old date or is this updating it with a new date/do you know which one it should be?

backend/src/app.ts
```diff
@@ -319,7 +319,48 @@ app.post('/api/add-like', authenticate, likeHandler(false));
...
+
+    // Send a success response with the updated review document ID
+    res.status(201).send(reviewDocId);
```
mluo24, line 339: status code 200 for updating

mluo24 (commented):
Nice work Cyrus! I have a few nits, but for the most part it looks great! Great use of the soft delete and validation!

backend/src/app.ts
```diff
@@ -319,7 +319,48 @@ app.post('/api/add-like', authenticate, likeHandler(false));
...
+    // Send a success response
+    res.status(200).send('Success');
+  } catch (err) {
+    // Handle any errors that may occur during the deletion process
+    console.log(err);
+    res.status(500).send('Error');
+  }
+});
+
+app.put(' :reviewDocId/:newStatus', async (req, res) => {
```
thuypham03, line 363: Please revert back to its original

thuypham03 (changes requested):

PR #327, opened 2023-11-20
Update Readme for FA23

thuypham03:


AnkitLakkapragada (approved):
Thank you for updating the Readme. It's important to keep this up to date so that people can refer to it whenever and get the most up to date information!

PR #332, opened 2023-12-01
[WIP] Edit review pop-up model

cyrus14:
### Summary 

This pull request is the first step towards implementing the edit review pop-up model within the Profile Page.

This is what has been done so far:

https://github.com/cornell-dti/cu-apts/assets/60114462/af152c17-6644-4d4e-b317-fe1adf5d2bed

I am still working on how to populate with existing review data. I have condensed the data into FormData which is correct format but for some reason the hearts are not changing even when being updated.

### Test Plan 

Tested out the various actions.

### DIFF
--- a/backend/src/app.ts
+++ b/backend/src/app.ts
@@ -1,3 +1,11 @@
+/**
+ * This file contains the backend logic for handling API requests related to reviews, landlords, apartments,
+ * and user interactions. It utilizes Express.js for routing and Firebase for data storage.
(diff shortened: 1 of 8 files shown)


AnkitLakkapragada (commented):
Great work Cyrus! I'm looking forward to this PR being finished as it is definitely needed and also will also improve the overall experience for the user since they can edit their own reviews. I really like the documentation that you have also added throughout the PR. Looking forward to it!

PR #336, opened 2023-12-02
Add Saved Property sort-by button for review count and average rating

mluo24:
### Summary 

This adds the ability to sort the saved apartments by either the number of reviews and the average rating of the apartment.

A lot of things had to change to make this work, including a bunch of async await calls, so loading times might kind of screwed.

### Test Plan 

image

image


ggsawatyanon (approved):
This works great and is really useful.

PR #345, opened 2024-02-19
Add Casper to README

CasperL1218:
Added name Casper into README contributors


kea-roy (changes requested):
Just remove the .zshrc file and then everything is good :)

PR #346, opened 2024-02-21
Added add_price_and_bedroom_attributes.ts file

kea-roy:
This script adds the price and bedrooms attributes to all the reviews that do not have them: setting them to zero as a default.

Tested by running on dev instance of FireStore.

### DIFF
--- a/backend/scripts/add_price_and_bedroom_attributes.ts
+++ b/backend/scripts/add_price_and_bedroom_attributes.ts
@@ -1,0 +1,19 @@
+import { Review } from '@common/types/db-types';
+import { db } from '../src/firebase-config';
+
+// This script will add the price and bedrooms attributes to all the reviews that do not have them
+
+const reviewCollection = db.collection('reviews');
+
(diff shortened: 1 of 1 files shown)


backend/scripts/add_price_and_bedroom_attributes.ts
```diff
@@ -0,0 +1,19 @@
+import { Review } from '@common/types/db-types';
+import { db } from '../src/firebase-config';
+
+// This script will add the price and bedrooms attributes to all the reviews that do not have them
```
ggsawatyanon, line 4: For the future, you should use a docstring/multi line documentation to document entire components or scripts. You should only use `//` for documenting specific lines of code. For example for this script you can write:
```
/** 
* This script will add the price and bedrooms attributes to all the reviews that do not have them
*/
```

ggsawatyanon (approved):
Good job on this script, Kea-Roy. This was really useful for adding price and bedroom attributes to our existing reviews.

PR #350, opened 2024-02-22
Implement Landlord Redirect Button (Change to Button Style)

CasperL1218:
Implement Landlord Redirect Button (Change to Button Style)

- [x] Change the landlord redirect text on the review from underlined text to a button style

Screenshot 2024-02-21 at 18 12 49
*The [Property: ] text is not underlined now

Screenshot 2024-02-21 at 18 12 53
*The text/button area becomes darker when the mouse hovers over it. It has a similar style to the [Visit Landlord] button on ApartmentCard.


kea-roy (changes requested):
Nice work! Works as intended with responsive styling. Tested on various screen sizes and the button looks correct for all sizes. 

But I don't think any of the styling is necessary. If you want to simplify the code a bit, you can try removing your custom stylings on lines 195-200. I tried running the code without the styling and it seems to look the same. Perhaps Material UI takes care of the styling for this component?

CasperL1218:
I tried removing the styling, and it works the same way as intended. 

I have removed the styling on lines 195-200.

kea-roy (dismissed):

ggsawatyanon (changes requested):
Good job on your first task! Good implementation of the button. Could you match the landlord button styling further by making the text not capitalized? Also ensure that the font size is the same as the review text font size.
Example below:
image

PR #352, opened 2024-02-24
Improved DropDown component to support new sorting features, Added optional price and bedrooms attributes to Review type, and Added avgRating and avgPrice attributes to CardData

kea-roy:
### Summary
This pull request is the first step towards implementing improved sorting features for apartment card lists, by...
1) Adjusting TypeScript typings to include price and bedroom fields for Reviews
2) Adjusting TypeScript typing to include avgRating and avgPrice for CardData
3) Improve DropDown component to support sorting on newer fields and look more like the figma design.

### Changes in More Depth

To be able to sort on average rating and price, the CardData type needs these additional fields so that the ApartmentCards element can access this data.

However, to calculate average rating and price, the Review type also needs additional fields of price and number of bedrooms. Thus, I have added these as optional attributes for now to avoid breaking any code.

To support new sorting options, I have also updated the DropDown.tsx file to be more flexible with menu options by removing the hard-coded default display value of recent and setting the default display value to be the first value in the menu instead.

I have also added an svg icon to the drop down button to match the figma design.

### Test Plan

Tested on different screen sizes, tested on different pages, including home page, location pages, and search results page.

### Notes

The actual sorting and front-end is completed, but will be added in separate pull request to make it easier to review.

### Breaking Changes/ Future Things to Consider

Since the Review datatype is used in many parts of the code, I've set the price and bedrooms attributes to optional for now. Once the updated review user interface is completed, the price and bedrooms attributes can be set to required.

### DIFF
--- a/backend/src/app.ts
+++ b/backend/src/app.ts
@@ -209,10 +209,21 @@ const pageData = async (buildings: ApartmentWithId[]) =>
 
       const numReviews = reviewList.docs.length;
       const company = landlordDoc.data()?.name;
+      // calculate average rating using overall rating of the reviews
+      const avgRating =
+        reviewList.docs.reduce((acc, curr) => acc + curr.data().overallRating, 0) /
+        Math.max(numReviews, 1);
+      // calculate average price using price category of the reviews, excluding reviews with price 0
+      const reviewsWithPrice = reviewList.docs.filter((review) => review.data().price > 0);
+      const avgPrice =
+        reviewsWithPrice.reduce((acc, curr) => acc + curr.data().price, 0) /
+        Math.max(reviewsWithPrice.length, 1);
       return {
         buildingData,
         numReviews,
         company,
+        avgRating,
+        avgPrice,
       };
     })
   );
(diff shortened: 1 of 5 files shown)


CasperL1218 (approved):
Nice work with the setup for the implementation. The choice of optional fields provides great flexibility. Aside from the sorting feature, I think you can look into the unused state of [isMobile]. I think it will be a good idea to discuss whether this is a design choice or an unnecessary field for these particular sections of the pages.

ggsawatyanon (changes requested):
Good job on your first task! Your changes to the types, average ratings/price, setting default value, and addition of the drop down arrow are great and very useful. However, the removal of isMobile has caused the border radius on mobile display to no longer be curved. You can return the previous code for isMobile or change it if you want to.
Desktop display (good):
image
Mobile display after your changes (no longer curved):
image
Mobile display before your changes (still curved):
image

kea-roy:
> Good job on your first task! Your changes to the types, average ratings/price, setting default value, and addition of the drop down arrow are great and very useful. However, the removal of isMobile has caused the border radius on mobile display to no longer be curved. You can return the previous code for isMobile or change it if you want to. Desktop display (good): image Mobile display after your changes (no longer curved): image Mobile display before your changes (still curved): image

## Commit Changes
1. Fixed styling on ApartmentPage.tsx when using the mobile version of the component.

ggsawatyanon (approved):
Good job on fixing the mobile version as I previously requested!

PR #367, opened 2024-09-20
[Ready for Review] Implemented Edit Reviews Feature

kea-roy:
### Summary 

It can be really frustrating as a user when you make a small error and have to make an entirely new review from scratch. 

With this feature, the edit reviews functionality streamlines the process and encourages users to refine their reviews, add photos, and update their reviews based on new experiences.

### Changes Made

This pull request makes the following changes:
- [x] [api endpoint for edit-review created, MoreVertIcon imported](https://github.com/cornell-dti/cu-apts/pull/367/commits/5df32d27eca15ecdc88ee119bdca814e4ca756e4)
- [x] [Improved edit-review API endpoint with more authentication checks, added review-by-id API endpoint](https://github.com/cornell-dti/cu-apts/pull/367/commits/d45ce6501c15f2699581e8490827a38b079a8b92)
- [x] [Added review autofill support to review modal, ensuring backward compatibility](https://github.com/cornell-dti/cu-apts/pull/367/commits/f8ddebccaabf99c9ef68fd4ad86469a80052553d)
- [x] [Changed ReviewRating and HeartRating to support a default value.](https://github.com/cornell-dti/cu-apts/pull/367/commits/866f04f430955ebe896c6b81262f47f5b4628182)
- [x] [Changed photo preview to support already uploaded images.](https://github.com/cornell-dti/cu-apts/pull/367/commits/c11b243d78e6b642eb8f9e1af937c86735a89428)
- [x] [changed uploadFile to not reupload files that have already been uploaded](https://github.com/cornell-dti/cu-apts/pull/367/commits/4d0b8ddcd5c1bd9fabf370292f6749a770aa1555)
- [x] [Added toast notification support, added review modal support from review, refreshes after edit changes, fixed typos, configured functionality for MoreVertIcon](https://github.com/cornell-dti/cu-apts/pull/367/commits/d4a85af7b59692d8c20737557b6daca5d9845147)
- [x] [Refactor ReviewComponent key prop assignment to ensure React refreshes the review components correctly](https://github.com/cornell-dti/cu-apts/pull/367/commits/5994658274203acb7117cc61f77c34cad97d3c03)

### Test Plan 

To test it thoroughly, I ran 15 "approval -> pending" tests, checking changes to each field separately. I made sure that the database was correctly updated and that the front-end user saw those changes reflected immediately in the UI.

**01: Start State**
01-start-state
**02: Prefills Correctly**
02-prefills
**03: Made Edits**
03-edits-made
[...]

PR #369, opened 2024-09-21
[Documentation] Added documentation to ReviewRating, including the defaultValue.

kea-roy:
### Summary 

Added documentation to ReviewRating component including a default value for the HeartRating component.

This pull request also implements defaultValue in the same way #367 has implemented it.


ggsawatyanon (approved):
Good job on the thorough documentation and including the types of the arguments and return.

PR #370, opened 2024-09-22
Remove old admins & Add new admins

ggsawatyanon:
### Summary 

- [x] Removed emails of admins who are no longer on the team
- [x] Added new admins

### DIFF
--- a/frontend/src/constants/HomeConsts.ts
+++ b/frontend/src/constants/HomeConsts.ts
@@ -1,17 +1,13 @@
 export const loadingLength = 5;
 export const admins: string[] = [
-  'cri23@cornell.edu', // Cyrus
-  'asl256@cornell.edu', // Ankit
   'gs622@cornell.edu', // Grace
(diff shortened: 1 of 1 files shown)


kea-roy:
## PR Review

I love how you cleanly removed the old admins and added the new admins. This is a pretty simple PR so there isn't much to critique. 

However, I have some suggestions for ways this list could be further improved.
1. It would be a good idea to be consistent with admin rights depending on role/subteam. For instance, add other developer leads to the admin list, given that Sophia maintains admin rights. 
2. Additionally, sorting the admin lists by each student's school year (senior, junior, etc..) may be more beneficial for easier addition and removal of names. 
3. This PR also reminds me of the need to move to a different admin system, where we track admins in the database instead of in a constant. Moving to such a system will provide admins more privacy and opportunities for easier management of admin rights in the future (ex: through a GUI).

Overall, this is a fabulously done PR and has no significant issues that need to be addressed.

kea-roy (approved):

PR #378, opened 2024-11-17
Converted save image button

ggsawatyanon:
### Summary 

This pull request fixes the bug where our save apartment button is an image and not an actual button as requested by PM.

- [x] Converted save apartment image button to normal button component

### Test Plan 
image
Apartment when saved:
image

### Notes 

- Currently, on mobile there is no way to actually save the apartment. I tried playing around with different locations to put a bookmark icon button but the space is quite cramped and the different versions I tried out made the UI too cluttered or too small, I'm going to ask designers about their opinions on where we can add the save button on mobile.

### DIFF
--- a/frontend/src/pages/ApartmentPage.tsx
+++ b/frontend/src/pages/ApartmentPage.tsx
@@ -39,8 +39,8 @@ import { getAverageRating } from '../utils/average';
 import { colors } from '../colors';
 import clsx from 'clsx';
 import { sortReviews } from '../utils/sortReviews';
-import savedIcon from '../assets/filled-large-saved-icon.png';
-import unsavedIcon from '../assets/unfilled-large-saved-icon.png';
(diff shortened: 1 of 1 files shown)


CasperL1218 (dismissed):
This is a simple and great fix to the historical issue; the implementation is clean and works well!

https://github.com/user-attachments/assets/aa36c2c3-8017-4fff-b2ca-c2dace1876e7

One small frontend enhancement: Currently the hover state of the saved button moves many components of the page as shown. It might be better to remove this effect unless this is an intentional design.

kea-roy:
## PR Changes
1. Resolved border resize issue when hovering.

kea-roy (approved):
This PR is great and the button's design works as intended. After resolving minor issues with material-ui styling, the button now works perfectly.

PR #386, opened 2025-01-15
Fixed Travel Time Document Scripts

CasperL1218:
### Summary 

This pull request is a fix of the travel time documents creation scripts. The scripts had a problem with not able to identify the request call host when it internally calls another endpoint. This pull request implements a dynamic variable that supplies that section of the base url.

- [x] Implemented dynamic base url for internal api calls

### Test Plan 

The tests for creating the actual travel time documents will be conducted soon.


ggsawatyanon (approved):
This fixed the issue with using the batch-create-travel-times api. I've now used this to update the prod database.

PR #390, opened 2025-02-10
WIP: Fixing Self-like Bug on Reviews

parsa-tehranipoor:
### Summary 

This pull request is a work in progress and attempts to fix a bug where users could like their own comment. This was fixed by adjusting the previously existing "likeHandler" function to include more requirements for a "like" on a comment to be performed. Unit tests still need to be performed for this PR.

- [x] Fixed the "likeHandler" method to stop users from liking their own reviews
- [ ] Created unit tests for "likeHandler" function

### Test Plan 

To test this PR out, two users are required. One user is required to create a comment and attempt to like it, while a separate user is required to like the same comment.
Screenshot 2025-02-09 at 11 49 53 PM
The first half of this test has been completed where a user can not like their own comment. However, a separate user is still needed to check if the like button still works.

### Breaking Changes  

### DIFF
--- a/frontend/src/components/Review/Review.tsx
+++ b/frontend/src/components/Review/Review.tsx
@@ -46,6 +46,7 @@ import getPriceRange from '../../utils/priceRange';
 import OptionMenu from '../utils/OptionMenu';
 import { ReactComponent as BedIcon } from '../../assets/bed-icon.svg';
 import { ReactComponent as MoneyIcon } from '../../assets/money-icon.svg';
+import { exists } from 'fs';
 
 type Props = {
   readonly review: ReviewWithId;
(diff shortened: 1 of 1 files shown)


kea-roy:
## PR Changes
- Removed exists import
- Cleared warnings from lint by adjusting == to ===

ggsawatyanon (changes requested):
This fix was effective in no longer allowing users to like their own review as requested in the bug. However, as the "Helpful" button is still active, when a user clicks the like button on their own review, the button still reacts to their click visually but does not increment the likes count. While this does fix the bug, it would be clearer for users if you could disable the button too for their own reviews – this would still display the button visually but when a user clicks the button, it does not visually react (i.e. turn gray). Good job on one of your first bug fixes on CUApts, though! Keep going :)

ggsawatyanon:
It seems the button is still not disabled (visually disabled) when liking your own review. See below for a video demonstration of the button still visually reacting to liking your own review. I would reccomend doing this by conditionally altering the CSS class or style depending on whether the review is your own review or not. We can discuss this in further details during work sesh! 

https://github.com/user-attachments/assets/361a1437-9802-41d9-89d2-8186cb4b3cb5

ggsawatyanon (approved):
As I reviewed before, this PR look good. I'm going to approve this to deploy the bug fixes but for the future we'll disable the button for users' own reviews

PR #414, opened 2025-12-08
Flag deprecated endpoints for saving apartments and landlords

laurenp-2:
### Summary 

This pull request is the first step towards deprecating old save/check endpoints and improving the UX for apartment saves.
-[x] Marked endpoints related to saving apartments and landlords (/api/check-saved-apartment, /api/check-saved-landlord) as deprecated in the backend code. These endpoints are still functional but will be replaced in future updates with more reliable logic.
-[x] Updated the frontend for the apartment page and cards so that the bookmark icon reflects hover state instead of the saved state; hovering over the icon now shows the “saved” icon visually, while the actual save/unsave logic is handled on click.

### DIFF
--- a/backend/src/app.ts
+++ b/backend/src/app.ts
@@ -826,6 +826,10 @@ app.put('/api/delete-review/:reviewId', authenticate, async (req, res) => {
 /**
  * Handles saving or removing saved apartments for a user in the database.
  *
+ * @deprecated This endpoint is deprecated and will be removed in future versions.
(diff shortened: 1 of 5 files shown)


parsa-tehranipoor:
This PR looks great! Really good that you added the deprecated tag on top of the already existing documentation for those routes. One thing I would add is that below some of the newly deprecated routes, there are some comments that attempt to describe the route itself. These comments would be better placed on top of the route with its correlated documentation, which would follow our codebase's coding conventions. Outside of that, great job!

PR #415, opened 2026-02-16
Add apartment tag CRUD endpoints

NigelTatem:
### Title

Apartment tags backend: create, attach/detach, and list

### Summary

This PR introduces backend support for tagging apartments. A new tags collection is added in Firestore, and apartments (stored in buildings) now optionally reference tags by ID through a tags array on the document. The goal was to add a reusable Tag entity and minimal endpoints to support creating tags and managing their association with apartments.

Tags are stored with both a name and a normalizedName (trimmed and lowercased) to enforce uniqueness at the application level. Creating a tag with the same normalized name returns the existing tag rather than inserting a duplicate. Apartments attach and detach tag IDs idempotently, meaning repeated operations do not create duplicates or cause errors.

The following endpoints were added:

POST /api/tags to create or return an existing tag

GET /api/apts/:id/tags to list resolved tag objects for an apartment

POST /api/apts/:id/tags/:tagId to attach a tag

DELETE /api/apts/:id/tags/:tagId to detach a tag

Reads tolerate dangling references by skipping missing tag documents rather than failing.

Emulator-compatible initialization was added for test runs so Firebase Admin can run without a service account when NODE_ENV=test or FIRESTORE_EMULATOR_HOST is set. A Tags test suite was added using the Firestore emulator and Supertest to verify tag creation/deduplication, idempotent attach/detach behavior, and correct listing. All backend tests pass locally with yarn workspace backend test.

This PR does not include frontend work, tag-based filtering, authentication changes, or transactional protections against concurrent writes. Tag uniqueness and tag-array updates are currently enforced at the application level rather than through Firestore transactions.

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -35,8 +35,11 @@
   "devDependencies": {
     "@types/cors": "^2.8.10",
     "@types/express": "^4.17.11",
+    "@types/jest": "^26.0.15",
     "@types/morgan": "^1.9.2",
     "@types/supertest": "^2.0.11",
-    "eslint-import-resolver-typescript": "^2.5.0"
+    "eslint-import-resolver-typescript": "^2.5.0",
+    "jest": "^26.6.0",
+    "ts-jest": "^26.5.3"
   }
 }
--- a/backend/src/app.ts
+++ b/backend/src/app.ts
@@ -12,6 +12,7 @@ import {
   LandlordWithLabel,
   ApartmentWithLabel,
   ApartmentWithId,
+  TagWithId,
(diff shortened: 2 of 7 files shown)


laurenp-2 (commented):
This is a really clean PR! The new tag API endpoints align well with the existing project structure and API patterns, and the code is very readable. One small suggestion would be to add more specific error handling in the API calls bc right now everything is wrapped in a catch that returns a general 400 error, so more granular responses could help with debugging and UX. Overall, this looks great. Can’t wait to see how the feature turns out!

PR #423, opened 2026-04-17
Fix cra frozen tsconfig

CasperL1218:
### Summary
  - Persist a one-line fix to `react-scripts@4.0.0` via `patch-package` so Heroku builds don't crash                
  - Remove redundant `import firebase from 'firebase/app'` in `ApartmentPage.tsx` that caused a TypeScript type     
  mismatch across the codebase                                                                                      
                                                                                                                    
  **Details:**                                                                                                      
  - `verifyTypeScriptSetup.js` in CRA assigns `appTsConfig = readTsConfig` (same reference), then passes
  `readTsConfig` through `immer` which freezes it. Any subsequent write to `appTsConfig.compilerOptions` crashes    
  with `TypeError: Cannot assign to read only property`. Fix: deep-copy via `JSON.parse(JSON.stringify(...))` before
   assigning.                                                                                                       
  - The patch was previously applied manually to `node_modules` (lost on every fresh install). `patch-package` + a
  `postinstall` script now re-applies it automatically after every `yarn install`, including on Heroku.             
   
  ### Test Plan                                                                                                     
  Deploy to `cuapts-staging` — build should succeed without the `TypeError: Cannot assign to read only property
  'jsx'` crash.                                                                                                     
   
  ### Notes                                                                                                         
  - The patch targets exactly `react-scripts@4.0.0`. If/when react-scripts is upgraded (Phase 3 dependency
  upgrades), this patch file should be deleted — `patch-package` will error loudly on install if the patch no longer
   applies, making it easy to catch.
  - `firebase/app` import removal in `ApartmentPage.tsx`: the import was only used for the `firebase.User` type     
  annotation. All other files in the codebase rely on the ambient Firebase namespace for this type — the explicit   
  default import created a `firebase.User` vs `firebase.default.User` split that TypeScript treated as incompatible.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -40,7 +40,8 @@
     "eject": "react-scripts eject",
     "lint": "eslint . --ext .ts --ext .tsx",
     "lint:fix": "yarn lint --fix",
-    "tsc": "tsc -p tsconfig.json"
+    "tsc": "tsc -p tsconfig.json",
+    "postinstall": "patch-package"
   },
   "proxy": "http://localhost:8080",
   "browserslist": {
@@ -56,6 +57,7 @@
     ]
   },
   "devDependencies": {
-    "@types/google-map-react": "^2.1.10"
+    "@types/google-map-react": "^2.1.10",
+    "patch-package": "^8.0.1"
   }
 }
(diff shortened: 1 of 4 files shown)


bryanshao07 (approved):
Looks good, solves two small issues, and makes it easier to run a migration.

